PR #25, opened 2018-11-05
Added back Runtime Manager API topic

valkyrie69:
So link wouldn't be broken on docs site. This is just a temp workaround.


mcocaro (changes requested):
could we add a "See Also" with links to CI/CD page?

mcocaro (approved):
LGTM

PR #37, opened 2018-11-29
DO NOT MERGE Docs 3526 app versioning

valkyrie69:
Please provide a technical review--feedback/corrections/anything

NOTE: The decision about *where* in the navigation this documentation will be has not been made yet.


modules/ROOT/pages/app-versioning.adoc
```diff
@@ -0,0 +1,76 @@
+= Application Versioning
+
+Applications deployed to Runtime Fabric have a versioning feature for the application's configuration. Versioning enables you to:
+
+* View the configuration change history of the application.
+* Roll back configurations to an earlier version in case the currently applied configuration has issues or is unstable.
+* Make updates to a deployed application's configuration, including:
+** The application ZIP file
```
mcocaro: maybe application artifact instead? to avoid calling out zip for mule 3 and jar for mule4

modules/ROOT/nav.adoc
```diff
@@ -7,6 +7,7 @@
  ** xref:deploying-to-pcf.adoc[Deploying to Pivotal Cloud Foundry]
 * xref:continuous-deployment.adoc[Continuous Deployment]
 * xref:managing-deployed-applications.adoc[Managing Deployed Applications]
+ ** xref:app-versioning.adoc[Application Versioning]
```
mcocaro: do you think its better to move this file to Runtime Fabric section? check this page in which it refers to deploying applications in RTF: https://docs.mulesoft.com/runtime-fabric/1.0/deploy-to-runtime-fabric

modules/ROOT/pages/app-versioning.adoc
```diff
@@ -0,0 +1,76 @@
...
+** Application version
```
mcocaro: is this the artifact version? if so, I would consider it as part of changing the application deployment artifact

modules/ROOT/pages/app-versioning.adoc
```diff
@@ -0,0 +1,76 @@
...
+** Deployment target settings, such as runtime version, number of replicas, and whether the application is running in cluster mode
+** Resource allocations such as number of cores and memory allocation
+** Ingress and property values
```
mcocaro: missing STATUS as a part of the versioning here

modules/ROOT/pages/app-versioning.adoc
```diff
@@ -0,0 +1,76 @@
...
+* Upload a new ZIP file.
+* Make changes to the deployment target and resource allocation.
+* Modify values for *Ingress* and *Properties*.
+. Once your changes are made, click *Apply Changes*. +
+At the top of the Settings page, a status bar appears.
++
+image:app-versioning-applying-config.png[]
+. Click *View Status* to view the status for each replica in the deployment. +
+Replica statuses include:
+* Started (green): The replica is receiving traffic.
```
mcocaro: should we instead update and link to this page? https://docs.mulesoft.com/runtime-manager/managing-deployed-applications#status-states

modules/ROOT/pages/app-versioning.adoc
```diff
@@ -0,0 +1,76 @@
...
+* Starting: The process is starting on the replica.
+* Deleting: The deletion of the replica is in progress.
+* Deleted: The replica has been deleted.
+* Stopping: The replica is in the middle of the stopping process.
+* Stopped: The replica has stopped (is not receiving traffic).
+* Failed (red): The desired configuration failed to deploy. +
+The last successfully deployed configuration is retained until the new (desired) configuration is successfully deployed. +
+New configurations are applied as "rolling updates." For example, a replica is created with the desired configuration, and once that configuration is successfully applied, the replica with the old configuration is deleted, and this process is repeated with each replica in the deployment. This model means there is zero downtime while updates are applied to the application.
```
mcocaro: can we include a note here to note that applications that cannot run multiple versions in parallel should consider this?

modules/ROOT/pages/app-versioning.adoc
```diff
@@ -0,0 +1,76 @@
...
++
+image::app-versioning-config-applied-success.png[]
++
+The application status, configuration identifier, date, and other information is updated in the status box.  The new configuration is displayed at the top of the *Config changes* list, along with a *Last successful* label to indicate that it's the last configuration successfully deployed.
++
+If the updates to the configuration fail to deploy, you get an "Desired configuration wasn't applied" message, and the Configuration status displays an "Update failed" label. The configuration is displayed in the *Config changes* history without a label.
++
+image::app-versioning-update-failed.png[]
++
+In cases of failure, the deployment is rolled back to the last successful version of the configuration, but the status of the configuration still displays "Updated failed."
```
mcocaro: there are scenarios in which k8s will not rollback new configurations, lets add here that it is on a best-intention basis but not guaranteed.

modules/ROOT/pages/app-versioning.adoc
```diff
@@ -0,0 +1,76 @@
+= Application Versioning
```
mcocaro: can we add a text to explain that what's being versioned is the Deployment (Application + Configuration) automatically by ARM?

mcocaro (changes requested):
Added some comments

modules/ROOT/nav.adoc
valkyrie69: This is what I mentioned--Steven and I are going to talk about the best placement for this with Sundip. I am thinking it will likely be moved to RTF.

modules/ROOT/pages/app-versioning.adoc
valkyrie69: I don't know about this.  The replica statuses seem to have parity with Kubernetes statuses and slightly different from the application statuses mentioned on the page you mention. We can talk about this.

modules/ROOT/pages/app-versioning.adoc
valkyrie69: Changed to "application package."

modules/ROOT/pages/app-versioning.adoc
valkyrie69: Changed the text. You will need to re-review.

modules/ROOT/pages/app-versioning.adoc
valkyrie69: Changed the text. You can re-review.

modules/ROOT/pages/app-versioning.adoc
valkyrie69: Added.

modules/ROOT/pages/app-versioning.adoc
valkyrie69: Added a note, but you will probably want to re-review. I'm not sure if it needs more context.

modules/ROOT/pages/app-versioning.adoc
mcocaro: agree on that - however, if I'm a user and want to search more on replica status, the least place I'd look is Application Versioning. would you agree?

modules/ROOT/pages/app-versioning.adoc
```diff
@@ -0,0 +1,82 @@
...
+* Starting: The process is starting on the replica.
+* Deleting: The deletion of the replica is in progress.
+* Deleted: The replica has been deleted.
+* Stopping: The replica is in the middle of the stopping process.
+* Stopped: The replica has stopped (is not receiving traffic).
+* Failed (red): The desired configuration failed to deploy. +
+The last successfully deployed configuration is retained until the new (desired) configuration is successfully deployed. +
+New configurations are applied as "rolling updates." For example, a replica is created with the desired configuration, and once that configuration is successfully applied, the replica with the old configuration is deleted, and this process is repeated with each replica in the deployment. This model means there is zero downtime while updates are applied to the application. +
+[NOTE]
+Applications can't run multiple versions in parallel.
```
PabloMR92: I would remove this one, since rolling update means that there are different versions running at the same time

modules/ROOT/pages/app-versioning.adoc
valkyrie69: I added a section to managing-deployed-applications#status-states for Replica statuses. But I am keeping the statuses in both places so the users don't have to click off the page to see the info about statuses. Especially if we decide the app versioning doc should be in Runtime Fabric (different product from Runtime Manager).

mcocaro: sounds good!

mcocaro (approved):
LGTM!

PR #56, opened 2019-01-10
Update cloudhub-fabric.adoc

atheedom:
Change the CH implementation of OSv2 to DynamoDB, as implied it is here https://github.com/mulesoft/object-store/blob/develop/ARCHITECTURE.md

### DIFF
--- a/modules/ROOT/pages/cloudhub-fabric.adoc
+++ b/modules/ROOT/pages/cloudhub-fabric.adoc
@@ -63,7 +63,7 @@ If your application is already deployed, you must redeploy it in order for your 
 
 Internally, worker scaling and VM queue load balancing is implemented using Amazon SQS.
 
-Object stores in your applications are also always stored persistently using Amazon SQS, even if you did not enable Persistent queues for your application.
+Object stores in your applications are also always stored persistently using DynamoDB, even if you did not enable Persistent queues for your application.
 
 HTTP load balancing is implemented by an internal reverse proxy server. Requests to the application (domain) URL +http://appname.cloudhub.io+ are automatically load balanced between all the application's Worker URLs.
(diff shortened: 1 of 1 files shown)


valkyrie69 (commented):
Amazon SQS is still used for worker scaling and VM queue load balancing? So the only implementation change is that now object stores are stored using DynamoDB?

Just want to verify that's correct.

nishantkakar:
I don't see how SQS is used for Worker Scaling. Also, Cloudhub ObjectStore is currently using MySQL to store objects in the ObjectStore not DynamoDB or SQS.

nishantkakar:
We have a plan to phase out OSv1 and move to OSv2 which uses DynamoDB. @steveCluster might be able to provide more information regarding OSv2

steveCluster:
Here are the docs for OSv2: https://docs.mulesoft.com/object-store/

We do use dynamodb and S3 in the backend, but we don't really advertise that to customers.

valkyrie69:
@steveCluster  I think the first sentence should be removed entirely. 
Then:

Object stores in your applications are also always stored persistently, even if you did not enable persistent queues for your application.

steveCluster:
George Hoffer is the docs guy from our team.  You should coordinate any changes with him.

Object stores are not always stored persistently.  It's a long conversation that I can go over with you in person.

valkyrie69:
@steveCluster  This is actually in my doc--not George's. It's not about the object store, it's in the context of CloudHub, and I'm fairly sure that in that context object stores in the app are stored persistently. I'll verify with the CH team. Thanks!

valkyrie69:
Not going to make update. Don't need to doc underlying technology and implementation.

PR #64, opened 2019-01-24
Added Note that replay is not supported for Mule 4

rodrigo-neira:
At the moment, replay feature is only available for Mule 3 applications. I added a note stating this as many users would expect to see the replay button on their mule 4 apps.

### DIFF
--- a/modules/ROOT/pages/insight.adoc
+++ b/modules/ROOT/pages/insight.adoc
@@ -182,9 +182,7 @@ After investigating the logs and resolving the issue that caused the transaction
 
 [NOTE]
 ====
-At the moment, the replay feature is only available for Mule 3 applications.
-
-Replay only works for flows that have inbound endpoints.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/insight.adoc
```diff
@@ -180,7 +180,12 @@ image::chexcmsg.png[CHExcMsg]
 
 After investigating the logs and resolving the issue that caused the transaction failure – for example, by purchasing more API capacity  – you can simply replay the transaction.
 
-*Note*: Replay only works for flows that have inbound endpoints.
+[NOTE]
+====
+At the moment, the replay feature is only available for Mule 3 applications.
```
valkyrie69: Change wording to:

Currently, the replay feature is supported only for Mule 3 applications and for flows that have inbound endpoints.

valkyrie69 (changes requested):
Please change wording. See inline comment.

valkyrie69 (approved):

PR #81, opened 2019-03-04
Update resolve-private-domains-vpc-task.adoc

gustavo-caracoglia:
Removed sentence which does not make sense since DNS configuration is tied to each VPC.


sebastiankorol (approved):
LGTM

PR #83, opened 2019-03-13
DOCS-1868 TLS namespace to API Manager example

kevintroller:
DOCS https://docs.mulesoft.com/runtime-manager/building-an-https-service#services-under-api-manager-proxies provides an example on HTTPS configuration.
The two following entries would need to be added to support TLS configuration: 

http://www.mulesoft.org/schema/mule/tls http://www.mulesoft.org/schema/mule/tls/current/mule-tls.xsd
xmlns:tls="http://www.mulesoft.org/schema/mule/tls"


fermujica (approved):
Looks good to me!

PR #101, opened 2019-04-24
Anypoint VPN supports ike v1 and v2

ahuynhms:


sjsheridan:
This PR is inaccurate and will be covered in the larger changes we are working on.

PR #122, opened 2019-05-24
add CLI instructions

Mulesoft-Mysti:
I did a little restructuring, too. 

Note that the top of the topic says you can deploy via CloudHub API and the Mule-Maven plugin, but there are no instructions. Please let me know if you want me to delete those two items, or leave them in. I can't add the missing content as I don't know where it is. It's not in the quick start that we are deleting.

The second commit removes links to the other topic that we added content from, and removes a useless tutorials page for CloudHub that wasn't in the nav doc. You don't need a whole page to call out a single tutorial. It's mentioned in the CloudHub docs where it is relevant.

There was a bit more link cleanup in last commits, including removing links to the next orphan to be removed.


modules/ROOT/pages/deploying-to-cloudhub.adoc
```diff
@@ -22,7 +22,34 @@ You can also deploy to CloudHub through:
 
 == Deploying an Application
```
valkyrie69: I would leave the CloudHub in the title because this is specific to deploying to CloudHub. There are other deployment topics for deploying to other targets.

modules/ROOT/pages/deploying-to-cloudhub.adoc
```diff
@@ -37,8 +64,58 @@ The Deploy Application page opens:
 +
 image::deploying-to-cloudhub-8524f.png[]
 
+=== Deploy from CloudHub CLI
+
+To deploy a Mule app from the CloudHub Command Line Utility (CLI):
+
+. Verify that https://www.mulesoft.com/platform/saas/cloudhub-ipaas-cloud-based-integration[CloudHub access] is enabled on Anypoint Platform.
+. If you do not already have access to the Anypoint-CLI command line tool, install it using xref:runtime-manager::anypoint-platform-cli.adoc#installation[Anypoint Platform CLI installation instructions].
```
valkyrie69: This topic is in the runtime-manager repo so you don't need runtime-manager in the xref. Also, we should probably point them to xref:anypoint-platform-cli2.adoc

modules/ROOT/pages/deploying-to-cloudhub.adoc
```diff
@@ -37,8 +64,58 @@ The Deploy Application page opens:
...
+. Log into your Anypoint Platform account from the command line:
+  .. Enter your username: `anypoint-cli --username="user"`. 
+  .. Enter your password.
+. Use the `runtime-mgr application deploy` command providing the *name* of the app and the *location* of the deployable archive (`.zip`) file on your file system, for example:
++
+----
+runtime-mgr cloudhub-application deploy myMuleApp /Users/exported-app-folder/my-mule-app.zip
+----
++
+. If there are no issues with the name, location or any optional parameters provided, you should see a table like this:
```
valkyrie69: No "should"--just "you see a table like this."

modules/ROOT/pages/deploying-to-cloudhub.adoc
```diff
@@ -37,8 +64,58 @@ The Deploy Application page opens:
...
+│ Workers                      │ 1 vCore * 1                            │
+----
+
++
+. Congratulations, your app is now deployed on CloudHub!
 
-=== From SandBox
++
+[TIP]
+. Applications that are deployed via the CLI can be viewed and managed on Runtime Manager like any other application.
```
valkyrie69, line 108: via and can be ? (I don't know the SF style convention for using via)

modules/ROOT/pages/deploying-to-cloudhub.adoc
```diff
@@ -37,8 +64,58 @@ The Deploy Application page opens:
...
+  .. If you then want to stop the app, on the command line type `runtime-mgr cloudhub-application stop <app-name>`.
+  .. To exit the command line tool, hit `ctrl + c` twice.
```
valkyrie69: Instead of "hit" use press?

modules/ROOT/pages/deploying-to-cloudhub.adoc
```diff
@@ -37,8 +64,58 @@ The Deploy Application page opens:
...
+. xref:runtime-manager::anypoint-platform-cli.adoc[Anypoint Platform CLI] contains a full list of commands.
```
valkyrie69: Same comment as above. Should be:
xref:anypoint-platform-cli2.adoc

valkyrie69 (approved):
Just a few minor things. Thank you for doing this!

modules/ROOT/pages/deploying-to-cloudhub.adoc
Mulesoft-Mysti: "via" has been in English since the 1700s. It's not a foreign language abbreviation like i.e. or e.g.
However, since writers often misuse the word, it's good to be cautious. This is correct because it does literally mean "by means of".

Mulesoft-Mysti:
Don't forget Valkyrie's edits in the next pass Mysti!

modules/ROOT/pages/deploying-to-cloudhub.adoc
Mulesoft-Mysti: fixed & tested locally

modules/ROOT/pages/deploying-to-cloudhub.adoc
Mulesoft-Mysti: fixed and tested locally

Mulesoft-Mysti:
https://www.mulesoft.org/jira/browse/DOCS-5101 for redirects.

PR #135, opened 2019-06-21
clarity on moving apps to another VPC

ahuynhms:
it is not clear to customers that they need to redeploy/restart the app to move it into a new VPC associated to that environment.

### DIFF
--- a/modules/ROOT/pages/vpc-provisioning-concept.adoc
+++ b/modules/ROOT/pages/vpc-provisioning-concept.adoc
@@ -79,6 +79,10 @@ You can create your Anypoint VPC inside any of the available CloudHub regions. +
 The recommended region to use might vary depending on how you connect to your Anypoint VPC. If you are using a VPN tunnel, you might want to choose the CloudHub region closest to your data center. +
 However, if you are peering with your private AWS VPC, you need to create your Anypoint VPC in the same AWS region.
 
+== FAQ: How to deploy or move apps into a VPC
+
+You associate a VPC to an environment and deploy apps into that same environment (e.g. Sandbox). If the region selected for your app and VPC are the same, it will deploy your app into that VPC. Otherwise, your app will be deployed into the MuleSoft shared VPC. If the VPC and environment association changes (e.g. you remove and add a different VPC to this environment), your apps will not automatically move into the new VPC. You will need to re-deploy/restart your apps so they move into your new VPC.
+
 == See Also
 
 * xref:to-request-vpc-connectivity.adoc[To Request VPC Connectivity to Your Network]
--- a/modules/ROOT/pages/vpc-provisioning-concept.adoc
+++ b/modules/ROOT/pages/vpc-provisioning-concept.adoc
@@ -44,7 +44,7 @@ An additional Anypoint VPC license is required for each remote location that nee
 [[faq_how_to_size_vpc]]
 == FAQ: How to Size my VPC
 
-When you create a an Anypoint VPC, the range of IP addresses for the network must be specified in the form of a Classless Inter-Domain Routing (CIDR) block, using CIDR notation. +
+When you create an Anypoint VPC, the range of IP addresses for the network must be specified in the form of a Classless Inter-Domain Routing (CIDR) block, using CIDR notation. +
 This address space is reserved for Mule workers, so it cannot overlap with any address space used in your data center if you want to peer it with your VPC.
 
 To calculate the proper sizing for your Anypoint VPC, you first need to understand that the number of dedicated IP addresses is not the same as the number of workers you have deployed. +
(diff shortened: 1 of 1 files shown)


sjsheridan (changes requested):
The requested update is not quite right. I suggest the following instead.

== FAQ: How to deploy or move applications to a VPC

MuleSoft applications are deployed to an environment. When the deployment region matches the VPC region:
- If the environment is already associated with a VPC at the time of application deployment, the application will reside inside the VPC. 
- If the environment is associated with a VPC after the application deployment, the application will not reside inside the VPC until it is restarted.
When the deployment region does not match the VPC region, the application will be deployed outside of the VPC. This means it will not have access to any connectivity methods configured for the VPC.

modules/ROOT/pages/vpc-provisioning-concept.adoc
```diff
@@ -79,6 +79,16 @@ You can create your Anypoint VPC inside any of the available CloudHub regions. +
 The recommended region to use might vary depending on how you connect to your Anypoint VPC. If you are using a VPN tunnel, you might want to choose the CloudHub region closest to your data center. +
 However, if you are peering with your private AWS VPC, you need to create your Anypoint VPC in the same AWS region.
 
+== FAQ: How to deploy or move applications to a VPC
```
kbostrom, line 82: Is it common for us to use FAQ headings?
Is it OK that this is not a question, in that case?

modules/ROOT/pages/vpc-provisioning-concept.adoc
```diff
@@ -79,6 +79,16 @@ You can create your Anypoint VPC inside any of the available CloudHub regions. +
...
+
+MuleSoft applications are deployed to an environment (dev, sandbox, and so on).
```
kbostrom: Consider spelling out "development" and providing one more environment type, if possible.

modules/ROOT/pages/vpc-provisioning-concept.adoc
```diff
@@ -79,6 +79,16 @@ You can create your Anypoint VPC inside any of the available CloudHub regions. +
 The recommended region to use might vary depending on how you connect to your Anypoint VPC. If you are using a VPN tunnel, you might want to choose the CloudHub region closest to your data center. +
 However, if you are peering with your private AWS VPC, you need to create your Anypoint VPC in the same AWS region.
 
+== FAQ: How to deploy or move applications to a VPC
+
+MuleSoft applications are deployed to an environment (dev, sandbox, and so on). 
+
+When the deployment region matches the VPC region:
```
kbostrom, line 86: Is "deployment region" the same as "environment"? If so, consider choosing the more precise of the two terms, in this context, and using it only, to avoid confusion.

modules/ROOT/pages/vpc-provisioning-concept.adoc
```diff
@@ -79,6 +79,16 @@ You can create your Anypoint VPC inside any of the available CloudHub regions. +
...
+
+* If the environment is already associated with a VPC at the time of application deployment, the app resides inside the VPC.
+* If the environment is associated with a VPC after the app deployment, the app will not reside inside the VPC until it is restarted.
```
kbostrom: Consider "will not reside" > "does not reside"

"...until it is restarted" > "until the app is restarted"

modules/ROOT/pages/vpc-provisioning-concept.adoc
```diff
@@ -79,6 +79,16 @@ You can create your Anypoint VPC inside any of the available CloudHub regions. +
...
+* When the deployment region does not match the VPC region, the app is deployed outside of the VPC. This means it will not have access to any connectivity methods configured for the VPC.
```
kbostrom: "This means that the app does not have access to...."

Can it be moved later? The FAQ title implies moving as well as deploying.

kbostrom (approved):

modules/ROOT/pages/vpc-provisioning-concept.adoc
valkyrie69: It's not really unless it's in an actual FAQ topic. But this is legacy content and there are already several headings like this. I think all the FAQs should be broken out into an FAQ topic, but that was beyond the scope of this PR.

modules/ROOT/pages/vpc-provisioning-concept.adoc
valkyrie69: No, they are two separate things.

hannanelson (approved):
Addressed Kim's edits.

hannanelson (approved):

hannanelson:
Hi @sjsheridan - Valkyrie added your suggestions, Kim edited, and I implemented the edits. Can you take a look? Thanks, Hanna

sjsheridan (changes requested):
The last paragraph starting "When the deployment region does not..." should not be a bullet point.

"Set the deployment region to the VPC region to deploy the app inside the VPC." should be removed, it is not quite correct - there is more to consider than just the region.

sjsheridan (approved):
Looks good, thanks!

PR #136, opened 2019-06-28
Update installing-and-configuring-runtime-manager-agent.adoc

dhimate:


mcocaro:
Can you please separate this change in a way that is evident for customers that those steps are only required if the intention is to be able to use the Mule Agent with Mutual TLS *and* be able to connect to the Anypoint Control Plane?

dhimate:
Hi @mcocaro, I am closing this. Could you please review https://github.com/mulesoft/docs-runtime-manager/pull/143

PR #138, opened 2019-07-01
FQDN requirement to resolve internal hostnames

gustavo-caracoglia:
JIRA: DOCS-5349


sjsheridan:
The only change here is this part:
(make sure to specify FQDNs, otherwise the VPC default resolver will take place).

This is not quite right. It should read:
(make sure your applications call the backend resources by FQDN).

gustavo-caracoglia:
Thanks @sjsheridan, just fixed it, let me know if any other changes are needed.

valkyrie69 (approved):

PR #143, opened 2019-07-10
Corrected as per feedback on pull request #136

dhimate:
@mcocaro as per feedback on pull request #136, I have created a separate section for runtime connected to the control plane with an option to setup two-way TLS for runtime manager.  I will withdraw the earlier pull request
@valkyrie69 FYI:


hannanelson (approved):
I'm going to merge this, then apply my changes.

PR #144, opened 2019-07-11
Update VPC discovery form

sjsheridan:
We need to update the VPC discovery form. I will attach the new version that should be used in place of the existing one.


sjsheridan:
[VPC-form-v9.4-template.xlsx](https://github.com/mulesoft/docs-runtime-manager/files/3382730/VPC-form-v9.4-template.xlsx)

hannanelson (approved):
Addressed Kim's edits to Sara's new content

PR #148, opened 2019-07-16
Added note about DEBUG; also lots of edits.

hannanelson:
Bug fix is on line 70-71.

### DIFF
--- a/modules/ROOT/pages/event-tracking.adoc
+++ b/modules/ROOT/pages/event-tracking.adoc
@@ -34,7 +34,7 @@ Mule provides these notification types:
 
 == Pushed Event Information
 
-For each notification, the Runtime Manager agent pushes the following information to the Event Tracking Service:
+For each notification, the Runtime Manager agent pushes the following information to the event tracking service:
 
 * `timestamp`
 +
@@ -70,7 +70,7 @@ Annotations for the component, if any
 
 The Runtime Manager agent stores event tracking notifications for Mule apps, and then streams them to an event tracking service. Depending on your network topology and the design of your Mule apps, copying event tracking notifications over the network from the Runtime Manager agent to the online event tracking service can significantly impact performance.
 
-You can reduce performance impact by specifying tracking levels at different granularities. You can specify a global default event tracking level, and then override it in named Mule apps. You can further override the event tracking level in specific named flows within a Mule app.
+You can reduce performance impact by specifying tracking levels at different granularities. You can specify a global event tracking level, and then override it in named Mule apps. You can further override the event tracking level in specific named flows within a Mule app.
 
 In addition to the event tracking level, you can use a message replay facility to store every message for replay. Although it might lead to duplicate message processing, the replay facility provides a mechanism for Operations to recover and replay failed transactions. As with tracking levels, you can configure the replay level globally, and then override the level for specific Mule apps, or flows within a Mule app.
 
@@ -78,10 +78,17 @@ In addition to the event tracking level, you can use a message replay facility t
 
 The available event tracking levels are:
 
-* `None` — Specifies no debugging.
-* `BE` — Tracks only events related to business events.
-* `Tracking` — Tracks `BE` level plus endpoint, exception strategy, and transaction-related notifications.
-* `Debug` — Specifies `Tracking` level plus additional component and message processor notifications.
+* `None`
++
+Specifies no debugging.
+* `BE`
++
+Tracks only events related to business events.
+* `Tracking`
++
+Tracks `BE` level plus endpoint, exception strategy, and transaction-related notifications.
+* `Debug`
++ Specifies `Tracking` level plus additional component and message processor notifications.
 +
 [IMPORTANT]
 Because setting event tracking to `Debug` can result in slow performance, use this tracking level for debugging only.
@@ -106,23 +113,18 @@ MULE TYPE is the `org.mule.api.context.notification.ServerNotification` subclass
 
 == Configuring Event Tracking
 
-You can configure the tracking level at the global level, the app level, or the flow level.
+You can configure the tracking level for all apps (global), a single app, or for a flow.
 
 * Global
-+ Specifies how to track all apps, and their flows, deployed to Mule runtime engine.
++
+Specifies how to track all apps, and their flows, deployed to Mule runtime engine.
 * App
 +
-Overrides the global default tracking level for the specific app.
+Overrides the global tracking level for the specific app.
 * Flow
 +
 Overrides the global or app tracking for the specific flow.
 
-To configure the global default tracking level, either:
-
-* Edit the configuration file `mule-agent.yml`.
-* Use the agent API to configure in real time.
-+
-See <<Configure Tracking Using the Agent API During Runtime>>.
 
 
 === Configure Event Tracking in mule-agent.yml
@@ -136,9 +138,9 @@ You can configure event tracking in `mule-agent.yml`:
 
 ==== Configure Event Tracking Globally for All Mule Applications
 
-The global default configuration specifies how to track all apps, and their flows, deployed to Mule runtime engine. You set the tracking configuration levels in the `services/mule.agent.tracking.service` section of the `mule-agent.yml` file.
+The global configuration specifies how to track all apps, and their flows, deployed to Mule runtime engine. You set the tracking configuration levels in the `services/mule.agent.tracking.service` section of the `mule-agent.yml` file.
 
-This example sets the global tracking level to `TRACKING`. See <<Event Tracking Levels>> for detailed descriptions of the tracking levels.
+This example sets the global tracking to `TRACKING`. See <<Event Tracking Levels>> for detailed descriptions of the tracking levels.
 
 In `mule-agent.yml`:
 [source,yaml,linenums]
@@ -150,9 +152,9 @@ services:
 
 ==== Configure Tracking for a Mule App
 
-You can override the global default tracking level by setting tracking levels per Mule app.
+You can override the global tracking configuration by setting tracking for a Mule app.
 
-The three Mule app level tracking areas are: `trackingLevel`, `replayLevel`, and `trackedFlows`.
+The Mule app tracking properties are: `trackingLevel`, `replayLevel`, and `trackedFlows`.
 
 * `trackingLevel`
 +
@@ -170,9 +172,9 @@ Specifies how to store each message through every flow of the Mule app for possi
 |`ALL` |  Stores messages for replay for every flow in the Mule app.
 |===
 
-This example sets the tracking level for the app `mule-flights` and its flow to `TRACKING`, and disables the message replay facility for this Mule app. The `mule-flights` app sends some notifications to the Event Tracking Service, but doesn't copy each received message. None of the flows in the `mule-flights` app can be replayed from the Event Tracking Service.
+This example sets the tracking level for the app `mule-flights` and its flow to `TRACKING`, and disables the message replay facility. The `mule-flights` app sends some notifications to the event tracking service, but doesn't copy each received message. None of the flows in the `mule-flights` app can be replayed from the event tracking service.
 
-This example also sets `globalTrackingLevel` to override the default value `NONE`, so other Mule apps are also tracked at the `BE` level.
+This example also sets `globalTrackingLevel` to `BE`, which overrides the default value of `NONE`. With this setting, other Mule apps are also tracked at the `BE` level.
 
 In `mule-agent.yml`:
 [source,yaml,linenums]
@@ -188,7 +190,7 @@ services:
 
 ==== Configure Tracking for Flows in a Mule App
 
-You can override global and app-level settings for flows in the specified Mule app by adding the `trackedFlows` element. This element specifies what to track for a flow, and how message replay behaves.
+You can override global and app settings for flows in the specified Mule app by adding the `trackedFlows` element. This element specifies what to track for a flow, and how message replay behaves.
 
 This table describes the tracked flows options:
 
@@ -201,7 +203,7 @@ This table describes the tracked flows options:
 |`payloadExcluded` |  If `replayLevel` is `SOURCED` or `ALL`, this determines whether to also store the message payload for replay, or to just store the message metadata.
 |===
 
-This example adds flow-level tracking configuration to the `mule-flights` Mule app configuration. The global and Mule app-level tracking use the default values (`NONE`).
+This example adds tracking to a flow in the `mule-flights` app. The global and Mule app tracking use the default values (`NONE`).
 
 In `mule-agent.yml`:
 [source,yaml,linenums]
@@ -217,9 +219,9 @@ services:
 ----
 
 
-=== Complex Tracking Example
+==== Complex Tracking Example
 
-This example sets tracking levels and settings at all three levels: global, Mule app, and flow.
+This example sets tracking levels: global, Mule app, and flow.
 
 Global:
 
@@ -243,7 +245,7 @@ App `mule-ticketing`:
 
 All other apps in the Mule instance, and their flows:
 
-* trackingLevel: `OFF` (the default)
+* trackingLevel: `NONE` (the default)
 
 In `mule-agent.yml`:
 [source,yaml,linenums]
--- a/modules/ROOT/pages/event-tracking.adoc
+++ b/modules/ROOT/pages/event-tracking.adoc
@@ -10,13 +10,13 @@ image:logo-server-active.png[link="/runtime-manager/deployment-strategies", titl
 image:logo-pcf-disabled.png[link="/runtime-manager/deployment-strategies", title="Pivotal Cloud Foundry"]
 
 
-Event tracking enables you to to view real-time information about messages being processed by a specific flow. You can view message payloads before, during, and after they are processed, as well as information about the message processors in the flow and message metadata.
+Event tracking enables you to to view real-time information about messages being processed by a specific flow. You can view message payloads before, during, and after they are processed, and you can view information about the message processors in the flow and message metadata.
 
-Event tracking provides visibility into business transactions and events on your Mule servers, enabling you to track and analyze the flow and disposition of particular messages. For example, you can use the data to perform root cause analysis on a failure within a business transaction, or to identify message processing bottlenecks.
+Event tracking enables you to view business transactions and events on your Mule servers so that you can track and analyze the flow and disposition of particular messages. For example, you can use the data to perform root cause analysis on a failure within a business transaction, or to identify message processing bottlenecks.
 
 == Notification Types
 
-Mule runtime engine provides notifications about execution of a flow's components.
+Mule runtime engine notifies you about flow component execution.
 
 image::event-tracking.png[event_tracking]
 
@@ -36,27 +36,43 @@ Mule provides these notification types:
 
 For each notification, the Runtime Manager agent pushes the following information to the Event Tracking Service:
 
-* `timestamp` — Timestamp for the event
-* `notificationType` — Event type, such as Message Processor Event, and Inbound Endpoint Event
-* `action` — Event action, such as Begin, End, Complete, and Rollback
-* `resourceIdentifier` — Name of the flow
-* `muleMessage` — If configured, pushes the serialized Mule message or an error message.
+* `timestamp`
++
+Timestamp for the event
+* `notificationType`
++
+Event type, such as Message Processor event or Inbound Endpoint event
+* `action`
++
+Event action, such as Begin, End, Complete, or Rollback
+* `resourceIdentifier`
++
+Name of the flow
+* `muleMessage`
++
+(If configured) Serialized Mule message or an error message.
 
 The Runtime Manager agent doesn't send this information:
 
 * Mule Message ID
-* Exception — Message and stack trace in case of an exception
-* Path — Component Mule path
-* Annotations — Annotations for the component, if any
+* Exception
++
+Message and stack trace in case of an exception
+* Path
++
+Component Mule path
+* Annotations
++
+Annotations for the component, if any
 * Application
 
 == Event Tracking Notifications
 
-The Runtime Manager agent stores event tracking notifications for Mule apps, then streams them to an Event Tracking Service. Depending on your network topology, and the design of your Mule apps, copying event tracking notifications over the network from the Runtime Manager agent to the online Event Tracking Service can significantly impact performance.
+The Runtime Manager agent stores event tracking notifications for Mule apps, and then streams them to an event tracking service. Depending on your network topology and the design of your Mule apps, copying event tracking notifications over the network from the Runtime Manager agent to the online event tracking service can significantly impact performance.
 
-You can reduce performance impact by specifying tracking levels at different granularities. You can specify a global default event tracking level, then override it in named Mule apps. You can further override the event tracking level in specific named flows within a Mule app.
+You can reduce performance impact by specifying tracking levels at different granularities. You can specify a global default event tracking level, and then override it in named Mule apps. You can further override the event tracking level in specific named flows within a Mule app.
 
-In addition to the event tracking level, you can use a message replay facility to store every message for later replay. Although it can potentially lead to duplicate message processing, the replay facility provides a mechanism for Ops to recover and replay failed transactions. As with tracking levels, you can configure the replay level globally, then override the level for specific Mule apps, or flows within a Mule app.
+In addition to the event tracking level, you can use a message replay facility to store every message for replay. Although it might lead to duplicate message processing, the replay facility provides a mechanism for Operations to recover and replay failed transactions. As with tracking levels, you can configure the replay level globally, and then override the level for specific Mule apps, or flows within a Mule app.
 
 == Event Tracking Levels
 
@@ -70,9 +86,9 @@ The available event tracking levels are:
 [IMPORTANT]
 Because setting event tracking to `Debug` can result in slow performance, use this tracking level for debugging only.
 
-This table lists the types of notifications active in each debugging level.
+The following table lists the types of notifications active in each debugging level.
 
-The MULE TYPE is the `org.mule.api.context.notification.ServerNotification` subclass (in the `org.mule.context.notification` package) that defines the notification type.
+MULE TYPE is the `org.mule.api.context.notification.ServerNotification` subclass (in the `org.mule.context.notification` package) that defines the notification type.
 
 [%header%autowidth.spread]
 |===
@@ -88,9 +104,18 @@ The MULE TYPE is the `org.mule.api.context.notification.ServerNotification` subc
 |Component Message | ComponentMessageNotifications |  |  |  |√
 |===
 
-== Configure Event Tracking
+== Configuring Event Tracking
 
-You can configure the tracking level globally, for an app, or for a flow.
+You can configure the tracking level at the global level, the app level, or the flow level.
+
+* Global
++ Specifies how to track all apps, and their flows, deployed to Mule runtime engine.
+* App
++
+Overrides the global default tracking level for the specific app.
+* Flow
++
+Overrides the global or app tracking for the specific flow.
 
 To configure the global default tracking level, either:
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/event-tracking.adoc
```diff
@@ -9,19 +9,18 @@ image:logo-hybrid-active.png[link="/runtime-manager/deployment-strategies", titl
 image:logo-server-active.png[link="/runtime-manager/deployment-strategies", title="Anypoint Platform Private Cloud Edition"]
 image:logo-pcf-disabled.png[link="/runtime-manager/deployment-strategies", title="Pivotal Cloud Foundry"]
 
-== Overview
 
-Event tracking allows you to to view real-time information about messages being processed by a specific flow. You can view message payloads before, during, and after they are processed, as well as information about the message processors in the flow and message metadata.
+Event tracking enables you to to view real-time information about messages being processed by a specific flow. You can view message payloads before, during, and after they are processed, as well as information about the message processors in the flow and message metadata.
```
kbostrom: Do we have to specify "payload" when documenting messages? Isn't that the default meaning, whereas the less common reference is to nonpayload info, or "metadata"?

...processed, as well as information about...."  >  "...processed, and you can view information about...."
or
"...processed, as well as message and flow metadata...." [parallelism; not sure how to word second half]

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -9,19 +9,18 @@ image:logo-hybrid-active.png[link="/runtime-manager/deployment-strategies", titl
...
 
-Event tracking provides you with visibility into business transactions and events on your Mule servers, allowing you to track and analyze the flow and disposition of particular messages. For example, you can use the data to do root cause analysis on a failure within a business transaction, or to identify message processing bottlenecks.
+Event tracking provides visibility into business transactions and events on your Mule servers, enabling you to track and analyze the flow and disposition of particular messages. For example, you can use the data to perform root cause analysis on a failure within a business transaction, or to identify message processing bottlenecks.
```
kbostrom: "enables you to view...so that you can track and analyze...."?

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -9,19 +9,18 @@ image:logo-hybrid-active.png[link="/runtime-manager/deployment-strategies", titl
...
 
 == Notification Types
 
-Mule provides notifications about execution of a flow's components as shown below.
+Mule runtime engine provides notifications about execution of a flow's components.
```
kbostrom: "notifies you about flow component execution"?

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -37,40 +36,43 @@ Available notification types:
 
 For each notification, the Runtime Manager agent pushes the following information to the Event Tracking Service:
 
-* `timestamp`: Timestamp for the event
-* `notificationType`: Event Type (i.e: Message Processor Event, Inbound Endpoint Event)
-* `action`: Event Action (i.e: Begin, End, Complete, Rollback)
-* `resourceIdentifier`: (i.e: flow name )
-* `muleMessage`: If configured, pushes the serialized Mule message; if this is not possible, it pushes an error message
+* `timestamp` — Timestamp for the event
```
kbostrom: Suggest reformatting definition-type lists as follows:
o timestamp
   Time of the event  [consider providing format: hours, minutes, seconds? clock time? date?]

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -37,40 +36,43 @@ Available notification types:
...
+* `notificationType` — Event type, such as Message Processor Event, and Inbound Endpoint Event
```
kbostrom: "such as Message Processor event or Inbound Endpoint event"

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -37,40 +36,43 @@ Available notification types:
...
+* `action` — Event action, such as Begin, End, Complete, and Rollback
```
kbostrom: "or Rollback"

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -37,40 +36,43 @@ Available notification types:
...
+* `resourceIdentifier` — Name of the flow
+* `muleMessage` — If configured, pushes the serialized Mule message or an error message.
```
kbostrom: Can this be reworded to be noun-based, so that it is parallel with its peers in the list?
"Serialized Mule message or error message, pushed [where?] if configured"? Ack, that does not sound right. I'm sure you can do better.

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -37,40 +36,43 @@ Available notification types:
...
 
-The following are not sent by the Runtime Manager agent:
+The Runtime Manager agent doesn't send this information:
 
 * Mule Message ID
-* Exception: The message and stack trace in case of an exception
-* Path: The component Mule path
-* Annotations: Annotations for the component, if any
+* Exception — Message and stack trace in case of an exception
```
kbostrom: ditto re. definition list format

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -37,40 +36,43 @@ Available notification types:
...
+* Path — Component Mule path
+* Annotations — Annotations for the component, if any
 * Application
 
-== Configuring Event Tracking Levels and Notification Content
+== Event Tracking Notifications
 
-Event tracking notifications are stored in the Runtime Manager Agent on behalf of Mule applications, then streamed to an Event Tracking Service. Depending on your network topology, and how your Mule applications are designed, there may be significant performance impacts to copy event tracking notifications over the network from the Runtime Manager Agent to the online Event Tracking Service.
+The Runtime Manager agent stores event tracking notifications for Mule apps, then streams them to an Event Tracking Service. Depending on your network topology, and the design of your Mule apps, copying event tracking notifications over the network from the Runtime Manager agent to the online Event Tracking Service can significantly impact performance.
```
kbostrom: GLOBALLY reword to not use ", then" as a conjunction. Replace with "and" or "and then" or ", and then," as appropriate.

"an event tracking service"  (Capitalize only if a proper noun.)

Remove comma between "topology" and "and."

Per first comment in this box, I now see "the" rather than "an" event tracking service. Is it a proper noun (one=the) or not (an=any of more than one)?

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -37,40 +36,43 @@ Available notification types:
...
 
-Because these notification can have performance implications, you can control tracking levels at various granularities.
+You can reduce performance impact by specifying tracking levels at different granularities. You can specify a global default event tracking level, then override it in named Mule apps. You can further override the event tracking level in specific named flows within a Mule app.
```
kbostrom: ditto GLOBAL re: ", then..."

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -37,40 +36,43 @@ Available notification types:
...
 
-An event tracking level can be specified globally, then overridden by specific named Mule applications. The event tracking level can be further overridden by specific named flows within a particular Mule application.
-
-Besides the event tracking level, a message replay facility is available to store every message for later replay. Although this can potentially lead to duplicate processing of messages, the replay facility may offer a mechanism for Ops to recover and replay failed transactions. Like the tracking levels, the replay level can be configured globally, then overridden granularly for specific Mule applications, then further overridden for specific flows within a Mule application.
+In addition to the event tracking level, you can use a message replay facility to store every message for later replay. Although it can potentially lead to duplicate message processing, the replay facility provides a mechanism for Ops to recover and replay failed transactions. As with tracking levels, you can configure the replay level globally, then override the level for specific Mule apps, or flows within a Mule app.
```
kbostrom: "later replay"  >  "replay"  ["re"=later]

"can potentially" >  "can"  ["might" might be better in this case]

"Ops" = ? "Operations"? As a proper noun?

"...globally, then override the level for specific Mule apps, or flows within a Mule app."
   >  "...globally and then override the level for specific Mule apps or flows within a Mule app."

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -37,40 +36,43 @@ Available notification types:
...
 
 == Event Tracking Levels
 
-Available event tracking levels are:
+The available event tracking levels are:
+
```
kbostrom, line 80: ditto re. format

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -37,40 +36,43 @@ Available notification types:
...
+* `None` — Specifies no debugging.
+* `BE` — Tracks only events related to business events.
+* `Tracking` — Tracks `BE` level plus endpoint, exception strategy, and transaction-related notifications.
+* `Debug` — Specifies `Tracking` level plus additional component and message processor notifications.
++
+[IMPORTANT]
```
kbostrom, line 93: Your discretion, but you could add this text to the definition and still have impact, without using a notice format.

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -37,40 +36,43 @@ Available notification types:
...
+Because setting event tracking to `Debug` can result in slow performance, use this tracking level for debugging only.
 
-* `None` - No debugging.
-* `BE` - Only events related to business events are tracked.
-* `Tracking` - `BE` level + endpoint, exception strategy, and transaction related notifications.
-* `Debug` - Tracking level + additional component and message processor notifications.
+This table lists the types of notifications active in each debugging level.
 
-The following table lists the types of notifications active in each debugging level. The MULE TYPE is the `org.mule.api.context.notification.ServerNotification` subclass (they are in the `org.mule.context.notification` package) that defines the notification type.
+The MULE TYPE is the `org.mule.api.context.notification.ServerNotification` subclass (in the `org.mule.context.notification` package) that defines the notification type.
```
kbostrom: Omit "The"

Because this text comes between the introduction and the table itself, I would consider using "The following" instead of "This."

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -86,27 +88,32 @@ The following table lists the types of notifications active in each debugging le
 |Component Message | ComponentMessageNotifications |  |  |  |√
 |===
 
+== Configure Event Tracking
+
+You can configure the tracking level globally, for an app, or for a flow.
```
kbostrom: I completely misread the previous sentence the first time, thinking it was globally for either apps or flows. Perhaps switch the order?

     "You can configure default event tracking at the application level, at the flow level, or globally."

Is "default" necessary?

Can we add something about why/when/under what circumstances our audience might choose one over the other?

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -86,27 +88,32 @@ The following table lists the types of notifications active in each debugging le
...
 
-== Configuring Event Tracking
+To configure the global default tracking level, either:
```
kbostrom: ==Configuring Event Tracking

===Configure Event Tracking in mule-agent.yml
==== Configure Event Tracking Globally for All Mule Applications
==== Configure Tracking for a Mule App
==== Configure Tracking for Flows in a Mule App

=== Complex Tracking Example

=== Configure Tracking Using the Agent API During Runtime
==== Retrieve the Current Configuration
==== Modify the Current Configuration

Okay. I had to do that so that I could see why we have a four-level hierarchy, and to be sure the titles are appropriate.

The introductory text under Configuring Event Tracking should set reader expectations for three subtopics, in this order: config in mule-agent, complex example, and config in API. So I would expect to read about the differences between using a config file vs. doing this in real time and why/when I would choose one over the other. I would also expect to see why I need an example for using the file but not in real time.

Then for the first subtopic (using the yaml file), the same applies. Introduce the topics included within and summarize them, including why/when a user chooses one over the other.

The second topic is an example, I assume for the yaml file technique. Why is it not part of that parent topic? It appears that this whole set is about event tracking using either YAML or API, and that YAML involves more considerations.

This might require some reorg.

So, the rest of my edit will be copyediting what is here now, without addressing structural considerations.

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -86,27 +88,32 @@ The following table lists the types of notifications active in each debugging le
...
 
-You can configure the tracking level globally, per application, or per flow. 
+* Edit the configuration file `mule-agent.yml`.
+* Use the agent API to configure in real time.
++
+See <<Configure Tracking Using the Agent API During Runtime>>.
```
kbostrom: See org. edit suggestion above.
Suggest mentioning the two techniques (YAML vs. API) and comparing why audience would use one of the other.

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -86,27 +88,32 @@ The following table lists the types of notifications active in each debugging le
...
 
-To configure the tracking level, edit the configuration file `mule-agent.yml`, or in real time, use the agent API.
 
-[NOTE]
-Performance may be affected depending on the event tracking level.
+=== Configure Event Tracking in mule-agent.yml
 
-=== Configuring Event Tracking in `mule-agent.yml`
+You can configure event tracking in `mule-agent.yml`:
```
kbostrom, line 132: Instead of just providing links, suggest mentioning the three scenarios and why/when an audience might use each one.

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -86,27 +88,32 @@ The following table lists the types of notifications active in each debugging le
...
 
-You can configure a global default configuration for how every Mule application should be tracked when deployed to the Mule runtime. You can also override the global setting per Mule application, and even per flow within a particular Mule application.
+* <<Configure Event Tracking Globally for All Mule Applications,Globally for all apps deployed to Mule>>
+* <<Configure Tracking for a Mule App,For a single Mule app>>
+* <<Configure Tracking for Flows in a Mule App,For flows in a Mule app>>
 
-==== Configuring Tracking Globally for Every Mule Application
 
-Tracking configuration levels are set in the `services/mule.agent.tracking.service` section of the `mule-agent.yml` file. You can manually edit this section to change tracking levels.
+==== Configure Event Tracking Globally for All Mule Applications
```
kbostrom, line 139: Why does global tracking appear before app and flow tracking? If this is the best practice method, state so and why.

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -86,27 +88,32 @@ The following table lists the types of notifications active in each debugging le
...
 
-The global configuration affects all the applications in the Mule instance, as well as their flows.
+The global default configuration specifies how to track all apps, and their flows, deployed to Mule runtime engine. Set the tracking configuration levels in the `services/mule.agent.tracking.service` section of the `mule-agent.yml` file.
```
kbostrom: Clarify how "default" comes in to all this. Is global the default way to track events? state that in the topic set intro, and tell why, compared to by app or by flow.

"global default" or "default global"?

Are the "tracking configuration levels" = app, flow, and global? Clarify use of "level." The terminology is a little inconsistent in this topic set, among "configuration," "default configuration," "global configuration," "global default configuration,"....

"Set" > "You set" or "You can set" or "are set," because we are not providing an instruction here so much as the concept of setting levels of configuration.

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -86,27 +88,32 @@ The following table lists the types of notifications active in each debugging le
...
 
-Here is a configuration example where the tracking level is set to `TRACKING`. See the table above for more detailed descriptions of these tracking levels.
+This example sets the tracking level to `TRACKING`. See <<Event Tracking Levels>> for detailed descriptions of the tracking levels.
```
kbostrom: "sets the global tracking level to..."?

Consider moving cross-ref to a See Also topic.

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -116,29 +123,33 @@ services:
  globalTrackingLevel: TRACKING
 ----
 
-[NOTE]
-If `globalTrackingLevel` is not specified, the default value is `NONE`.
+If you don't specify `globalTrackingLevel`, the default value is `NONE`.
```
kbostrom: Why do we have to make a note of this? The example must be followed in all cases, right? So why document someone not following it?

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -116,29 +123,33 @@ services:
...
 
-==== Configuring Tracking per Mule Application
+==== Configure Tracking for a Mule App
 
-In addition to setting a default global tracking level for every application, you can override the tracking levels per Mule application.
+You can override the default global tracking level by setting tracking levels per Mule app.
```
kbostrom: Clarify "default": A default, to me, is something that exists before a user configures it. But the previous section implies that it must be configured, so...instead of overriding a default, isn't the user setting a new default? Is this really about removing an app from the global group and handling it differently? Is the missing point here that users set up global tracking, but can then isolate apps and flows later?

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -116,29 +123,33 @@ services:
...
 
-There are three Mule application level tracking areas: trackingLevel, replayLevel, and trackedFlows.
+The three Mule app level tracking areas are: `trackingLevel`, `replayLevel`, and `trackedFlows`.
```
kbostrom: app-level

What is a tracking "area"? This gets confusing because we refer to three tracking levels (global, app, flow), and now the app level has its own three sublevels? Can we make "level" more distinct?

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -116,29 +123,33 @@ services:
...
 
-* `trackingLevel` specifies one of the four tracking levels (`NONE`, `BE`, `TRACKING`, `DEBUG`) described in the previous table. This tracking level is applied to every flow in the Mule application.
-* `replayLevel` specifies if and how each message through every flow of the Mule application should be stored for possible later replay. The possible values are: `NONE`, `SOURCED`, or `ALL`.
+* `trackingLevel`
++
+Specifies one of the four tracking levels: `NONE`, `BE`, `TRACKING`, or `DEBUG`. This tracking level affects every flow in the Mule app. See <<Event Tracking Levels>>.
+* `replayLevel`'
++
+Specifies how to store each message through every flow of the Mule app for possible later replay: `NONE`, `SOURCED`, or `ALL`.
```
kbostrom: ditto re. "later replay"

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -116,29 +123,33 @@ services:
...
 
+[[Replay]]
 [%header%autowidth.spread]
 |===
 |Replay Level Option | Description
-|`NONE` |  Do not store any data related to message replay and disable the replay feature for these flows.
-|`SOURCED` | Only stores messages for later replay for each flow in the Mule application that starts with a message source.
-|`ALL` |  Stores messages for later replay for every flow in the Mule application.
+|`NONE` |  Specifies to not store any data related to message replay and to disable the replay feature for these flows.
+|`SOURCED` | Stores messages for later replay for only those flows in the Mule app that start with a message source.
```
kbostrom: ditto

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -116,29 +123,33 @@ services:
...
+|`ALL` |  Stores messages for later replay for every flow in the Mule app.
```
kbostrom: ditto; GLOBALLY search for and evaluate

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -116,29 +123,33 @@ services:
...
 |===
 
-In the following example, the tracking level for the application `mule-flights` and its flow is set to `TRACKING`, and the message replay facility is disabled for this Mule application. The `mule-flights` application will send some notifications to the Event Tracking Service, but will not copy each received message, and none of the flows in the `mule-flights` application can be replayed from the Event Tracking Service.
+This example sets the tracking level for the app `mule-flights` and its flow to `TRACKING`, and disables the message replay facility for this Mule app. The `mule-flights` app sends some notifications to the Event Tracking Service, but doesn't copy each received message. None of the flows in the `mule-flights` app can be replayed from the Event Tracking Service.
```
kbostrom: Consider omitting "for this Mule app."

ditto ? re. capitalization for Event Tracking Service

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -151,26 +162,23 @@ services:
       trackingLevel: TRACKING
       replayLevel: NONE
 ----
-In addition to changing the `trackingLevel` and `replayLevel` for a Mule application, you can also change these settings for flows inside the listed Mule application by adding the `trackedFlows` value.
 
-The `trackedFlows` element contains additional elements to specify granular tuning of what should be tracked for a flow, and how any message replay should behave. When a flow is configured, these settings override the settings inherited by the global and Mule application level settings.
+==== Configure Tracking for Flows in a Mule App
 
-==== Configuring Tracking per Flow of a Mule Application
+You can override global and app-level settings for flows in the specified Mule app by adding the `trackedFlows` element. This element specifies what to track for a flow, and how message replay behaves.
```
kbostrom: ditto previous comment about clarifying defaults and overrides

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -151,26 +162,23 @@ services:
...
 [%header%autowidth.spread]
 |===
 |Tracked Flow Option | Description
-|`flow name` |  The name of the flow, as it is coded in the Mule application's XML configuration file.
-|`tracking level` |  One of the values `NONE`, `BE`, `TRACKING`, or `DEBUG`, as described in the previous table.
-|`replayLevel` |  One of the values `NONE`, `SOURCED` (only store messages if the flow begins with a message source), `ALL`
+|`flow name` |  The name of the flow, as it is coded in the Mule app XML configuration file.
+|`tracking level` |  One of the values `NONE`, `BE`, `TRACKING`, or `DEBUG`. See <<Event Tracking Levels>>.
+|`replayLevel` |  One of the values `NONE`, `SOURCED`, or `ALL`. See the <<Replay,Replay Levels>> table.
 |`payloadExcluded` |  If `replayLevel` is `SOURCED` or `ALL`, this determines whether to also store the message payload for later replay, or to just store the message metadata.
```
kbostrom: Can we reword to make this entry parallel with the previous?

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -151,26 +162,23 @@ services:
...
+This table describes the tracked flows options:
 
 [%header%autowidth.spread]
 |===
 |Tracked Flow Option | Description
-|`flow name` |  The name of the flow, as it is coded in the Mule application's XML configuration file.
-|`tracking level` |  One of the values `NONE`, `BE`, `TRACKING`, or `DEBUG`, as described in the previous table.
-|`replayLevel` |  One of the values `NONE`, `SOURCED` (only store messages if the flow begins with a message source), `ALL`
+|`flow name` |  The name of the flow, as it is coded in the Mule app XML configuration file.
+|`tracking level` |  One of the values `NONE`, `BE`, `TRACKING`, or `DEBUG`. See <<Event Tracking Levels>>.
```
kbostrom, line 201: Suggest moving cross-ref

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -151,26 +162,23 @@ services:
...
+|`replayLevel` |  One of the values `NONE`, `SOURCED`, or `ALL`. See the <<Replay,Replay Levels>> table.
```
kbostrom, line 202: ditto: See Also

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -151,26 +162,23 @@ services:
...
 |`payloadExcluded` |  If `replayLevel` is `SOURCED` or `ALL`, this determines whether to also store the message payload for later replay, or to just store the message metadata.
 |===
 
-Here is an example that adds some flow level tracking configuration to the previous `mule-flights` Mule application configuration. The global and Mule application level tracking is using the default values (`NONE`).
+This example adds flow-level tracking configuration to the `mule-flights` Mule app configuration. The global and Mule app-level tracking use the default values (`NONE`).
```
kbostrom: "This example adds flow-level tracking to 'mule-flights.'"

"...use the default values (`NONE`)."    Maybe I am lost, but...was the default value set to NONE in a previous example, or is it true that the actual default value for all tracking is NONE?

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -186,29 +194,33 @@ services:
 ----
 
 
+=== Complex Tracking Example
+
+This example sets tracking levels and settings at all three levels: global, Mule app, and flow.
```
kbostrom: ditto confusion about two types of "levels"
Also, we have two uses of "set" and "settings" as well. Yikes.

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -186,29 +194,33 @@ services:
...
 
-=== Setting Complex Tracking Configurations
+Global:
 
-In the example below, tracking levels and settings are overridden at all three levels: global, Mule application, and flow.
+* `globalTrackingLevel` set to `OFF`, which is the default setting if you don't specify `globalTrackingLevel`.
```
kbostrom: See Line 26, which states that the default is NONE, not OFF....
This applies to all of the text that follows.

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -186,29 +194,33 @@ services:
 ----
 
 
+=== Complex Tracking Example
+
+This example sets tracking levels and settings at all three levels: global, Mule app, and flow.
 
-=== Setting Complex Tracking Configurations
+Global:
```
kbostrom, line 226: The narrative begins to look a bit like shorthand at this point. Will this make sense to our audience?

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -236,13 +248,14 @@ services:
             replayLevel: ALL
 ----
 
-=== Configuring Via the Agent API During Runtime
+=== Configure Tracking Using the Agent API During Runtime
+
+The agent API enables you to change the agent tracking configuration during runtime.
```
kbostrom: Can we add more context? Why/under what circumstances does a user want or need to do this?

Should these be task topics instead of concepts?

kbostrom (approved):
Very clean! But we might want to consider some reorganization at some point....

modules/ROOT/pages/event-tracking.adoc
hannanelson: Because it sets the default that you override.

modules/ROOT/pages/event-tracking.adoc
```diff
@@ -37,40 +36,59 @@ Available notification types:
...
+You can reduce performance impact by specifying tracking levels at different granularities. You can specify a global default event tracking level, and then override it in named Mule apps. You can further override the event tracking level in specific named flows within a Mule app.
 
-Besides the event tracking level, a message replay facility is available to store every message for later replay. Although this can potentially lead to duplicate processing of messages, the replay facility may offer a mechanism for Ops to recover and replay failed transactions. Like the tracking levels, the replay level can be configured globally, then overridden granularly for specific Mule applications, then further overridden for specific flows within a Mule application.
+In addition to the event tracking level, you can use a message replay facility to store every message for replay. Although it might lead to duplicate message processing, the replay facility provides a mechanism for Operations to recover and replay failed transactions. As with tracking levels, you can configure the replay level globally, and then override the level for specific Mule apps, or flows within a Mule app.
```
valkyrie69: Operations doesn't need to be capped

valkyrie69 (approved):

modules/ROOT/pages/event-tracking.adoc
hannanelson: Query in to @mcocaro.

hannanelson:
This is waiting for info on NONE vs OFF.

PR #152, opened 2019-07-17
Revert "Revert "Update VPC discovery form""

hannanelson:
Reverts mulesoft/docs-runtime-manager#151


hannanelson:
This PR is waiting for the location of the updated file so we can link from the topic.

valkyrie69 (approved):

PR #155, opened 2019-07-19
Remove the specific commands as 2nd level headers.

fermujica:
Also fixing broken anchor references


hannanelson (approved):
Made one tiny fix.

fermujica:
Thanks for catching that! 327538d

PR #160, opened 2019-07-24
Including required steps to enable VPN feature

thiagopri:
Including additional steps to enable VPN feature under the Anypoint Platform account.


hannanelson:
@mcocaro - please verify. Thanks.

hannanelson:
@sjsheridan confirms that these steps are unnecessary for normal flow. Only relevant if creating a VPC and VPN in a business group, which this doesn't apply to.

PR #161, opened 2019-07-24
Docs 5553 add steps for vpn

hannanelson:
@nishantkakar - please review.


hannanelson:
@sjsheridan confirms that these steps are unnecessary for normal flow. Only relevant if creating a VPC and VPN in a business group, which this doesn't apply to.

PR #168, opened 2019-08-07
Typo correction

comikem:
ClourHub -> CloudHub


valkyrie69:
Fixed in a different PR

PR #203, opened 2019-09-26
Added admonition.

hannanelson:
RE: https://mulesoft.slack.com/archives/C02HGAD2E/p1568939129013700

image

### DIFF
--- a/modules/ROOT/pages/lb-mapping-rules.adoc
+++ b/modules/ROOT/pages/lb-mapping-rules.adoc
@@ -14,10 +14,10 @@ By using proxy rules, you can map a domain or subdomain to one of your Mule appl
 
 == Using Patterns in Mapping Rules
 
-A pattern is a string that defines a template for matching an input text. Whatever is placed into curly brackets (`{   }`) is treated as a variable.
-These variables can contain only letters (a-z) and cannot contain any other characters, such as digits, slashes, etc. The variable values can contain the characters`a-z0-1.&?-_`, but can't contain slashes.
-
 IMPORTANT: Patterns in the application URI are not currently supported.
+
+A pattern is a string that defines a template for matching an input text. Whatever value is placed within curly brackets (`{   }`) is treated as a variable.
+Variable names can contain only lowercase letters (a-z) and no other characters. Variable values can contain the characters`a-z0-1.&?-_`, but can't contain slashes.
 
 Suppose you set up a DNS CNAME record from  `example.com` to `_lb-name_.lb.anypointdns.net`. You can then map `app.example.com` to a different deployed CloudHub Mule application name.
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -17,6 +17,8 @@ By using proxy rules, you can map a domain or subdomain to one of your Mule appl
 A pattern is a string that defines a template for matching an input text. Whatever is placed into curly brackets (`{   }`) is treated as a variable.
```
kbostrom:
```suggestion
A pattern is a string that defines a template for matching an input text. Whatever value is placed within curly brackets (`{   }`) is treated as a variable.
```

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -17,6 +17,8 @@ By using proxy rules, you can map a domain or subdomain to one of your Mule appl
...
 These variables can contain only letters (a-z) and cannot contain any other characters, such as digits, slashes, etc. The variable values can contain the characters`a-z0-1.&?-_`, but can't contain slashes.
```
kbostrom: Only lowercase letters, or both lowercase and uppercase?

kbostrom: What is the difference between a variable, which can contain only letters, and a variable value, which can contain [the string? or do you mean a-z, 0 or 1, ., $, 7, -, and _?] `a-z0-1.$7-_`?

kbostrom:
```suggestion
These variables can contain only letters (a-z). The variable values can contain the characters`a-z0-1.&?-_`, but can't contain slashes.
```
or
```suggestion
These variables can contain only letters (a-z) and no other characters, such as digits, slashes, and so on. The variable values can contain the characters`a-z0-1.&?-_`, but can't contain slashes.
```

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -17,6 +17,8 @@ By using proxy rules, you can map a domain or subdomain to one of your Mule appl
...
 
+IMPORTANT: Patterns in the application URI are not currently supported.
+
 Suppose you set up a DNS CNAME record from  `example.com` to `_lb-name_.lb.anypointdns.net`. You can then map `app.example.com` to a different deployed CloudHub Mule application name.
```
kbostrom, line 22: Is it clear to your audience how the second sentence follows logically from the first? It does not to me, because I don't know how the values in the first relate to the different values in the second, but I am not your audience....

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -40,7 +42,6 @@ For `input=”bookings.example.com”`, the pattern can be resolved by assigning
 
 Depending on your design, you can choose to leverage your internal redirects to your endpoints using patterns or simply literal mappings.
```
kbostrom:
```suggestion
Depending on your design, you can leverage your internal redirects to your endpoints by using either patterns or  literal mappings.
```

kbostrom (approved):
Here you go!

hannanelson:
@ahuynhms - I'm going to go ahead and merge. Let me know if you have any issues with the change and I'll update.

PR #222, opened 2019-11-06
Update vpn-about.adoc

sjsheridan:
Adding [[supported_settings]] target for reference from another location.


sjsheridan:
this is not actually needed - thank you Hanna!

PR #238, opened 2019-11-27
Moved to CloudHub section, uncommented icons.

hannanelson:


jorgegarciamule (changes requested):
I seeing now that moving the [Building an HTTPS Service] page is not just moving it. I see 2 sections inside [Mule Runtime] that look similar to [Building an HTTPS Service]. One is [Mule App Development Tutorial] and the other is [Build an Application from an API]. On both, we explain how to use things like APIKit or connectors. [Building an HTTPS Service] is on that category but it also talks about CloudHub at the beginning, which I recommend to remove to make it more generic, and also API Manager Proxies that I'm not sure if it make sense there.

hannanelson:
I guess I misunderstood your suggestion. Are you saying this is not CloudHub specific, @jorgegarciamule?

If this is meant to be generic, then how generic? Should it even be in the Runtime Manager/CloudHub content? Is the material covered elsewhere?

Can you give me an idea of specifically what changes you want? If you want me to remove the CH info, then where should this topic live?

jorgegarciamule:
Hi Hanna. There are so many things in this simple page that it took me some time to think where should it be, and I'm not still 100% sure.
I recommend: 
1) Moving the [Building an HTTPS Service] article to Mule Runtime, right after [Build an Application from an API]
2) Removing the Cloudhub references or make it broader explaining that this applies to any deployment model. There is a specific section in CloudHub article [CloudHub Networking Guide] covering this.
Let me know if it makes sense for you.

hannanelson:
Created bug DOCS-6709 to track move to Mule Runtime.

jorgegarciamule:
We are moving forward with this modification and push the big changes to another Jira

jorgegarciamule (approved):
Looks good

paigehargrave (approved):

PR #245, opened 2019-12-05
Added note and links for statistics storage.

hannanelson:
Updated https://docs.mulesoft.com/runtime-manager/monitoring-dashboards

### DIFF
--- a/modules/ROOT/pages/monitoring-dashboards.adoc
+++ b/modules/ROOT/pages/monitoring-dashboards.adoc
@@ -289,7 +289,7 @@ From this dashboard, you can click a <<Access the Server Monitoring Dashboard,se
 == See Also
 
 * xref:monitoring::index.adoc[Anypoint Monitoring Overview]
-* https://anypoint.mulesoft.com/exchange/portals/anypoint-platform/f1e97bc6-315a-4490-82a7-23abe036327a.anypoint-platform/cloudhub-api/1.0.11/console/method/%231708/[CloudHub API - dashboardStats]
+* https://anypoint.mulesoft.com/exchange/portals/anypoint-platform/f1e97bc6-315a-4490-82a7-23abe036327a.anypoint-platform/cloudhub-api/1.0.11/console/method/%231708/[CloudHub API dashboardStats]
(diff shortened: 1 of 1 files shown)


mjmule (approved):
looks good. Thank you.

modules/ROOT/pages/monitoring-dashboards.adoc
```diff
@@ -38,6 +38,10 @@ The app monitoring dashboards help you answer questions about your integrated ap
 * What impact do high traffic peaks have on the message fail rate?
 * What impact do high traffic peaks of other apps running on the same server have on the performance of a given app?
 
+
+[NOTE]
+After the worker is restarted, the CPU and memory information is not available in the dashboards. To retain these statistics, use the `dashboardStats` API. Alternatively, you can use Anypoint Monitoring with a Titanium subscription to store this data.
```
kbostrom, line 43: Will it be clear to your audience (or does it even matter) whether this is restarted/current information that is not available or historic/cumulative information? 
In the context of the previous text, focused on information available from the dashboards, I wonder if we should mention that before the condition of restarting a worker (a corner case), maybe using parallel wording instead of a Note?
```suggestion
The app monitoring dashboards cannot provide CPU and memory information if the worker is restarted. In this case, you either retain this information by using the `dashboardStats` API or by storing it using Anypoint Monitoring with a Titanium subscription.
```

modules/ROOT/pages/monitoring-dashboards.adoc
```diff
@@ -284,6 +288,8 @@ From this dashboard, you can click a <<Access the Server Monitoring Dashboard,se
 
 == See Also
 
+* xref:monitoring::index.adoc[Anypoint Monitoring Overview]
+* https://anypoint.mulesoft.com/exchange/portals/anypoint-platform/f1e97bc6-315a-4490-82a7-23abe036327a.anypoint-platform/cloudhub-api/1.0.11/console/method/%231708/[CloudHub API - dashboardStats]
```
kbostrom: Consider reworking topic titles so that they do not use punctuation (CloudHub API - dashboardStats).

kbostrom (approved):
This looks just fine.
My comments are brainstoming-level only.

PR #248, opened 2019-12-10
Documentation - Runtime Manager, RAMLs and Cloudhub API ex…

shambhuprasad:
…change

### DIFF
--- a/modules/ROOT/pages/lb-create-arm.adoc
+++ b/modules/ROOT/pages/lb-create-arm.adoc
@@ -36,7 +36,7 @@ Each dedicated load balancer exposes an external domain `_lb-name_.lb.anypointdn
 
 . Select a target Anypoint VPC from the drop-down list.
 
-. Sets the Mule Application Response Timeout value. The default value is 300.
+. Set the Mule Application Response Timeout value. The default value is 300.
 
 . Add any whitelisted classless inter-domain routing (CIDR) as required.
 +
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/lb-create-arm.adoc
```diff
@@ -36,6 +36,8 @@ Each dedicated load balancer exposes an external domain `_lb-name_.lb.anypointdn
 
 . Select a target Anypoint VPC from the drop-down list.
 
+. Sets the Mule Application Response Timeout value. The default value is 300.
```
hannanelson: I do not see this in the Create Load Balancer UI.

nishantkakar (approved):

modules/ROOT/pages/lb-create-arm.adoc
shambhuprasad: Yes. The UI is not released yet in production. However, you can check the UI in our stgx environment. We need to merge this PR after UI goes in production.

hannanelson: @shambhuprasad - What is the schedule for the change? This kind of UI change will likely affect screenshots - I need to schedule this work. Thanks.

hannanelson: @shambhuprasad - can we change the label from "Timeout in Seconds" (which doesn't say what the timeout is for) to "Application Timeout" or "Mule Application Timeout"? 

What exactly is this option for? Is it the time to wait for a Mule app before....doing what?

shambhuprasad: UI change is deployed in production as hotfix as it was escalated. Please review "Load Balancer create/update" to see the updated changes.

vinayakjangam (approved):

sarathecoubian (approved):

PR #255, opened 2019-12-18
Docs 3837 feedback need help with uri params in output url hn

hannanelson:
Revised topic. Also fixed some little errors I found along the way. 
@sjsheridan, @henrymollman  - can you take a look?

Original: https://docs.mulesoft.com/runtime-manager/lb-mapping-rules

See attached PDF for rewrite.

[Dedicated Load Balancer Mapping Rules _ MuleSoft Documentation.pdf](https://github.com/mulesoft/docs-runtime-manager/files/4321778/Dedicated.Load.Balancer.Mapping.Rules._.MuleSoft.Documentation.pdf)


modules/ROOT/pages/lb-create-arm.adoc
```diff
@@ -53,8 +53,17 @@ Accepts the inbound request on the default SSL endpoint using the HTTP protocol.
...
 Redirects the request to the same URL using the HTTPS protocol.
+. Specify options:
++
+* *Disable Static IPs* specifies to use dynamic IPs, which do not persist when the DLB restarts.
+* *Keep URL encoding* specifies to encode the space (`%20`) and hash (`%23`) special characters.
++ 
+[IMPORTANT]
+The DLB only encodes the `%20` and `%23` characters. If a requesting URL includes other special characters, they are not encoded and result in errors, such as `Illegal character in path`. See https://help.mulesoft.com/s/article/Some-Special-Characters-Do-Not-Work-in-URLs-to-DLB[DLB and Special Characters].
+* *Support TLS 1.0* supports Transport Layer Security (TLS) v1 encryption in addition to TLS v1.1 and TLS v1.2.
+* *Upstream TLS 1.2* enables TLS v1.2 encryption on upstream traffic.
```
henrymollman: You may want to specify that this option will enable tls v1.2 to the mule workers behind the DLB, it may not be clear to the user.

hannanelson: Thanks @henrymollman - that wasn't clear to me! Updated.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,128 +8,262 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
 image:logo-rtf-disabled.png[link="/runtime-manager/deployment-strategies", title="Runtime Fabric"]
 
-The load balancer configuration is defined by a list of mapping rules that describe how an input URL should be translated into calls to different CloudHub apps.
-Mapping rules are attributes of the load balancer's SSL endpoint.
-When you create a mapping rule, you need to specify a certificate CN. Omitting the `[certificateName]` parameter adds the mappings to the default endpoint.
 
-When you create a simple matching rule, one input address is matched to the defined output: the endpoint of one of your applications.
-Instead of using literal matchings, you can also use a pattern to match a variable, such as input text to an endpoint.
+The CloudHub dedicated load balancer (DLB) routes requests from clients to Mule apps deployed within the VPC.
+_Mapping rules_ enable you to rename requests to the DLB (input URL) to a different Mule application name and domain.
```
danieldiaz-mule: "Rename" is confusing, it implies URL rewrites happening, which may not be the case. I'd say something along the lines of "enables you to forward requests to the DLB to a different..."

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,128 +8,262 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
 
-By using proxy rules, you can map a domain or subdomain to one of your Mule applications running in CloudHub.
+You can either define mapping rules when you create the DLB or add them to an existing DLB using Runtime Manager, the command-line interface (CLI), or the CloudHub API.
```
danieldiaz-mule: We may want to call out somewhere that if no rules are defined a default rule will be created.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,128 +8,262 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
 
-== Using Patterns in Mapping Rules
+When you create a simple matching rule, one input address matches the defined output: the endpoint an application.
```
danieldiaz-mule: This is just confusing. Does it mean to read "the endpoint of an application"?

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,128 +8,262 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+You can also use a pattern to match a variable, such as input text to an endpoint.
+
+By using proxy rules, you can map a domain or subdomain to a Mule application running in CloudHub.
```
danieldiaz-mule: What are proxy rules? There's no place in the ui to define "Proxy rules", only mapping rules.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,128 +8,262 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
-Instead of using literal matchings, you can also use a pattern to match a variable, such as input text to an endpoint.
+The CloudHub dedicated load balancer (DLB) routes requests from clients to Mule apps deployed within the VPC.
+_Mapping rules_ enable you to rename requests to the DLB (input URL) to a different Mule application name and domain.
 
-By using proxy rules, you can map a domain or subdomain to one of your Mule applications running in CloudHub.
+You can either define mapping rules when you create the DLB or add them to an existing DLB using Runtime Manager, the command-line interface (CLI), or the CloudHub API.
 
-== Using Patterns in Mapping Rules
+When you create a simple matching rule, one input address matches the defined output: the endpoint an application.
+You can also use a pattern to match a variable, such as input text to an endpoint.
```
danieldiaz-mule: which pattern? input URI pattern? App Pattern?

Maybe also add a note about patterns in the appURI not being supported.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,128 +8,262 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
-IMPORTANT: Patterns in the application URI are not currently supported.
+== Create Mapping Rules
 
-A pattern is a string that defines a template for matching an input text. Whatever value is placed within curly brackets (`{   }`) is treated as a variable.
-Variable names can contain only lowercase letters (a-z) and no other characters. Variable values can contain the characters `a-z0-1.&?-_`, but can't contain slashes.
+Mapping rules are attributes of the load balancer's SSL endpoint, which is identified by the certificate Common Name (CN).
+When you create mapping rules with:
 
-Suppose you set up a DNS CNAME record from  `example.com` to `_lb-name_.lb.anypointdns.net`. You can then map `app.example.com` to a different deployed CloudHub Mule application name.
+* Runtime Manager, you define mapping rules within a certificate definition.
```
danieldiaz-mule: What is a "Certificate definition"? You're talking about SSL endpoints and then move to Certificate Definitions

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,128 +8,262 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
-Suppose you set up a DNS CNAME record from  `example.com` to `_lb-name_.lb.anypointdns.net`. You can then map `app.example.com` to a different deployed CloudHub Mule application name.
+* Runtime Manager, you define mapping rules within a certificate definition. 
+* The CLI or CloudHub API, you must specify a certificate CN.
++
+If you omit the certificate name, the mappings are added to the default endpoint.
+
+=== Create Mapping Rules in Runtime Manager 
 
-For example, you can bind two hostnames for redirect:
+To create mapping rules in the certificate for a DLB in Runtime Manager:
```
danieldiaz-mule: You don't create mapping rules in a certificate (SSL endpoint instead)

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,128 +8,262 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+image::cloudhub-dlb-mapping-rules.png[the *Add New Rule* icon and the fields required to create a mapping rule]
+. Complete the fields:
++
+* *Input Path* (`inputUri`)
++ 
+The URL that the client requests: for example, `/{app}/`
++
+[IMPORTANT]
+_What does this mean:_
+The input URL follows the main load balancer’s domain and should remain constant for the same load balancer.
```
danieldiaz-mule: This is confusing. The main load balancer's domain doesn't make sense. This should be the contents of the "Host" header.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,128 +8,262 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
++
+The name of the CloudHub application that processes the request: for example, `{app}-example`.
+* *Output Path* (`appURI`)
++
+URI string to pass to the app: for example, `/`.
++
+The *Output Path* cannot contain patterns.
+* *Protocol* (`upstreamProtocol`)
++
+The protocol on which the application listens: `http` (port 8091), `https` (8092), `ws` (WebSockets), or `wss` (Web Service Security).
```
danieldiaz-mule: Websocket uses the same ports as HTTP:
http = ws = 8091
https = wss = 8092

modules/ROOT/pages/lb-mapping-rules.adoc
danieldiaz-mule: A URL is composed of several portions:
Schema: https://
Host: mydomain.com
Port: :443 (most of the times omitted if it's the default port for the schema - http 80/https 443)
URI Stem: /apis/myapiname/endpoint
Query string: ?blah=blah&meh=90

The "main load balancer's domain" is not either of these, although it refers to the hostname. In this case makes more sense to refer to the host header.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,128 +8,262 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
++
+By default, the load balancer listens to external requests on HTTPS and communicates internally with your worker over HTTP. If you configured your Mule application within the VPC to listen on HTTPS, make sure you set *Protocol* to `https` when you create the mapping rule list.
+
+You can use either literal strings or patterns to define the input and output path.
+
+[IMPORTANT]
+_Remove this?_
+Depending on your design, you can leverage internal redirects to your endpoints by using either patterns or literal mappings.
```
danieldiaz-mule: Well... it's not really relevant. But if you decide to keep it, you can say internal redirects to your application endpoints...

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,278 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+----
+/organizations/{orgid}/loadbalancers
+----
 
-=== URL Mapping
 
-You can pass the app name as an input URI and map it directly to the app name in CloudHub:
+== Using a Literal 1:1 Mapping
+
+If you have only one application, you can map the literal app name.
```
danieldiaz-mule: Remove or put it as an example. It currently read as it would be possible to do ONLY if you have 1 input URI mapped to an application. 

You can also do literal 1:1 mappings for multiple applications by changing the InputURI to be static instead of a defined pattern:
0 /v1/applicationX    application-x http
1 /v1/applicationY    application-y http
2 /v2/applicationx    application-x-v2

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,278 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+If you set up a DNS CNAME record to map `example.com` to `_lb-name_.lb.anypointdns.net`,
+you can then use a mapping rule to map `app.example.com` to a different deployed CloudHub Mule application name.
+
+[%header,cols="10a,20a,20a,20a,10a"]
 |===
-|Rule # |Input URL 2+^| Output URL
-|   |  *URI*   |       *appName*    |  *appURI*
-| 0 | /        | {subdomain}   | /
+| Rule # | *Input Path* | *Target App* | *Output Path* | *Protocol*
+|1 | /api/v1/ | default-app | /api/v1/ | http
```
danieldiaz-mule: Thiss is going to catch traffic to second-app, too. It'll pass it to default-app/api/v1/second-app/<uri-params>

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,278 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+|2 | /api/v1/secondapp | second-app | /api/v1	http
```
danieldiaz-mule: Don't know why but when I change to rich diff mode instead of source diff, I can't see this rule

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,278 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
-* Passing `application.example.com` is mapped to `application.cloudhub.io`.
 
-The same applies to the https://en.wikipedia.org/wiki/Subject_Alternative_Name[Subject Alternative Names] (SANs) of your SSL endpoints.
+You can route external requests to `+https://app.example.com+` to `+http://app.cloudhub.io/example+`:
 
-Instead of using a wildcard certificate, which permits any subdomain application name to be passed through the dedicated load balancer, you can restrict the allowed subdomain names. To do this, you must have different SANs configured for a certificate's common name. You can use the `subdomain` variable to map the subdomain portion of your domain name to your application.
+[%header,cols="10a,20a,20a,20a,10a"]
+|===
+| Rule # | *Input Path* | *Target App* | *Output Path* | *Protocol*
+|1 | app.example.com | app | /example | http
```
danieldiaz-mule, line 128: Makes sense... haven't really tested it but it was documented as if that would work and looking at the regex it should. I'll have to validate it later on (but it looks ok based off on what was written before)

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,278 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
-[%header,cols="10a,20a,20a,10a"]
+Alternatively, you can define a pattern to hold the input value:
+
+[%header,cols="10a,20a,20a,20a,10a"]
 |===
-|Rule # |Input URL 2+^| Output URL
-|   |  *URI*   |       *appName*    |  *appURI*
-| 0 | /        | {subdomain}-app   | /
+| Rule # | *Input Path* | *Target App* | *Output Path* | *Protocol*
+|1 | example.com/{myPattern} | app-{myPattern} | /data | http
```
danieldiaz-mule, line 136: |1 | /{myPattern}/ | app-{myPattern} | /data | http

Where did you get this off? It's repeated several times. I don't know if it's the view that's showing extra stuff or not.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: This is in the existing documentation: https://docs.mulesoft.com/runtime-manager/lb-mapping-rules

Removed.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: This example came from the current docs: https://docs.mulesoft.com/runtime-manager/lb-mapping-rules#using-patterns-in-mapping-rules:

> For example, you can bind two hostnames for redirect:
> ‘app.example.com’ ->  application: `app` URI: `/example’
> In this case, external requests to https://app.example.com are directed to http://app.cloudhub.io/example.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: This is from the existing docs:
> Or you can define a pattern to hold the input value:
> ‘example.com/{mypattern}’ -> application: `app-{mypattern}` URI: /data
> The example above causes that both example.com/bookings and example.com/sales match to app-bookings/data and app-sales/data respectively, as the variable mypattern holds these values. For input=”bookings.example.com”, the pattern can be resolved by assigning mypattern=”bookings” and for input=`sales.example.com`, the pattern is resolved to assign mypattern=”sales”.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: Good catch! There's a bug in the table (missing pipe on the second rule before http). Fixed.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: I got this example from this KB article: https://help.mulesoft.com/s/article/Check-list-of-using-DLB-mapping-rules
> Take a real case for an example, with the rules below, URL "/api/v1/secondapp" will be mapped to the app "default-app" rather "second-app". In order to pick the more exact mapping, you need to put the second rule on top of the first.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: Removed.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: In the DLB UI, mapping rules are defined within the Certificate definition: DLB>Add Certificate>URL Mapping Rules

What do you suggest instead?

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: This is from the existing doc:
> When you create a simple matching rule, one input address is matched to the defined output: the endpoint of one of your applications. Instead of using literal matchings, you can also use a pattern to match a variable, such as input text to an endpoint.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: This is from the existing doc and made no sense to me either:
>The input URL follows the main load balancer’s domain and should remain constant for the same load balancer.

hannanelson: Changed to "The input URL is appended to the host header of the load balancer."

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: Added back "Instead of using literal matchings," 

This is the intro - I don't think we want to talk about the details and limitations of pattern matching here. It's covered in the next section.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: Not sure what you mean by repeated several times - I'm not seeing that in the source.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: OK, take a look now and let me know if I messed up :)

modules/ROOT/pages/lb-mapping-rules.adoc
danieldiaz-mule: I know... we can leave it as is and change it when I have it tested.

modules/ROOT/pages/lb-mapping-rules.adoc
danieldiaz-mule: Looks like it's gone now... previously when loading the diff view it was multiplicating entire paragraphs.

danieldiaz-mule (approved):

modules/ROOT/pages/lb-mapping-rules.adoc
danieldiaz-mule: Within a SSL endpoint identified by its corresponding certificate's Common Name (CN)

modules/ROOT/pages/lb-mapping-rules.adoc
danieldiaz-mule: I don't see a difference
`In this case, the rule must match the application exactly. 

This rule maps the default load balancer `lb-demo.lb.anypointdns.net` directly to the `myApp.cloudhub.io` app in CloudHub:

[%header,cols="10a,20a,20a,20a,10a"]
|===
| Rule # | *Input Path* | *Target App* | *Output Path* | *Protocol*
| 1 | / | myApp | / | http 
|===

This rule matches the `dev-misc-mythellotest-v1-myapp` app:`

modules/ROOT/pages/lb-mapping-rules.adoc
danieldiaz-mule: > 3. Pay attention to the sequence of the rules. DLB always picks up the first matching rule regardless if there are more exact matchings later. Take a real case for an example, with the rules below, URL "/api/v1/secondapp" will be mapped to the app "default-app" rather "second-app". In order to pick the more exact mapping, you need to put the second rule on top of the first.

It's the "bad" example

danieldiaz-mule (changes requested):

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: It used to say:
>If you have only one application, you can map the literal app name.
It now says:
> Instead of using a pattern, you can map the literal app name.
> In this case, the rule must match the application exactly.

If this is incorrect, please let me know how can I correct it.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: Yes, I added another table:

 In the following example, the `/api/v1/secondapp` URL maps to the `default-app` app:

[%header,cols="10a,20a,20a,20a,10a"]
|===
|Priority|Input Path	|Target App| Output Path| Protocol
|1 | /api/v1/ | default-app | /api/v1/ | http
|2 | /api/v1/secondapp | second-app | /api/v1 |	http
|===

For the rule to match only `second-app`, the rule priority must be:

[%header,cols="10a,20a,20a,20a,10a"]
|===
|Priority|Input Path	|Target App| Output Path| Protocol
|1 | /api/v1/secondapp | second-app | /api/v1 |	http
|2 | /api/v1/ | default-app | /api/v1/ | http
|===

danieldiaz-mule: Looks better, just get /api/v1/secondapp/ with that trailing slash

modules/ROOT/pages/lb-mapping-rules.adoc
danieldiaz-mule: I'd do 

> Instead of using a pattern, you can map literal app names.
In this case, the rules must match the applications exactly.

Again, not a single app for the static mapping (1:1). You can do a single app by mapping / but that's not the same as saying you can do 1:1 mappings.

hannanelson: OK, replaced.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: Added.

danieldiaz-mule (approved):

modules/ROOT/pages/anypoint-platform-cli-commands.adoc
```diff
@@ -203,7 +203,7 @@ In order for the Anypoint Platform CLI to recognize your target servers, each se
 |<<runtime-mgr serverGroup remove server>> | Removes server from a server group
 |===
 
-== Commands for Administrating your Local Cluster Servers
```
kbostrom, line 206: lol

modules/ROOT/pages/anypoint-platform-cli-commands.adoc
```diff
@@ -229,7 +229,7 @@ In order for the Anypoint Platform CLI to recognize your target servers, each se
 |<<runtime-mgr standalone-alert list>> | Lists all alerts for standalone runtimes in the environment
 |===
 
-== Commands for Managing your CloudHub Dedicated Load balancer
+== Commands for Managing your CloudHub Dedicated Load Balancer
```
kbostrom:
```suggestion
== Commands for Managing Your CloudHub Dedicated Load Balancer
```
If you use the capitalization tool, be sure to click the Chicago tab.

modules/ROOT/pages/anypoint-platform-cli-commands.adoc
```diff
@@ -249,7 +249,7 @@ In order for the Anypoint Platform CLI to recognize your target servers, each se
 |<<cloudhub load-balancer whitelist remove>>| Remove an IP or range of IPs from the Load balancer whitelist
 |<<cloudhub load-balancer mappings describe>>| Lists the proxy mapping rules for a Load balancer. If no `certificateName` is given, the mappings for the default SSL endpoint are shown
 |<<cloudhub load-balancer mappings add>>| Add a proxy mapping rule at the specified index. If no `certificateName` is given, the mappings for the default SSL endpoint are shown
-|<<cloudhub load-balancer mappings remove>>| Remove a proxy mapping ruleIf no `certificateName` is given, the mappings for the default SSL endpoint are shown
+|<<cloudhub load-balancer mappings remove>>| Remove a proxy mapping rule. If no `certificateName` is given, the mappings for the default SSL endpoint are shown
```
kbostrom: period
```suggestion
|<<cloudhub load-balancer mappings remove>>| Remove a proxy mapping rule. If no `certificateName` is given, the mappings for the default SSL endpoint are shown.
```

modules/ROOT/pages/lb-create-arm.adoc
```diff
@@ -97,3 +97,4 @@ Click *Add New Rule*, and then specify the input path, target app, output path,
 * xref:lb-whitelists.adoc[Whitelists]
 * xref:lb-mapping-rules.adoc[Mapping Rules]
 * xref:lb-create-cli.adoc[Create a Load Balancer (Anypoint Platform CLI)]
+* xref:lb-create-api.adoc[Create a Load Balancer (CloudHub API)]
```
kbostrom: Our style is to avoid punctuation or formatting in titles; could you reword/reorder a bit to comply?

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
-Mapping rules are attributes of the load balancer's SSL endpoint.
-When you create a mapping rule, you need to specify a certificate CN. Omitting the `[certificateName]` parameter adds the mappings to the default endpoint.
 
-When you create a simple matching rule, one input address is matched to the defined output: the endpoint of one of your applications.
-Instead of using literal matchings, you can also use a pattern to match a variable, such as input text to an endpoint.
+The CloudHub dedicated load balancer (DLB) routes requests from clients to Mule apps deployed within the VPC.
+_Mapping rules_ enable you to forward requests to the DLB (input URL) to a different Mule application name and domain.
 
-By using proxy rules, you can map a domain or subdomain to one of your Mule applications running in CloudHub.
+You can either define mapping rules when you create the DLB or add them to an existing DLB using Runtime Manager, the command-line interface (CLI), or the CloudHub API.
```
kbostrom: Do you think we should specify "...the DLB, or you can define and add them to an existing ...later, using...."

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
 
-== Using Patterns in Mapping Rules
+When you create a simple matching rule, one input address matches the defined output: the endpoint of an application.
+Instead of using literal matchings, you can use a pattern to match a variable, such as input text to an endpoint.
+
+The order of the rules determines the priority. The first rule in the list that matches is applied.
```
kbostrom:
```suggestion
The order of the rules determines their priority. The first rule in the list that matches is applied. 
```

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+You can either define mapping rules when you create the DLB or add them to an existing DLB using Runtime Manager, the command-line interface (CLI), or the CloudHub API.
 
-== Using Patterns in Mapping Rules
+When you create a simple matching rule, one input address matches the defined output: the endpoint of an application.
+Instead of using literal matchings, you can use a pattern to match a variable, such as input text to an endpoint.
+
+The order of the rules determines the priority. The first rule in the list that matches is applied. 
+
+[NOTE]
+If you don't define any mapping rules, the DLB creates and uses a default mapping rule.
```
kbostrom, line 24: Should we add that you can overwrite that later?

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
 
-A pattern is a string that defines a template for matching an input text. Whatever value is placed within curly brackets (`{   }`) is treated as a variable.
-Variable names can contain only lowercase letters (a-z) and no other characters. Variable values can contain the characters `a-z0-1.&?-_`, but can't contain slashes.
+== Create Mapping Rules
+
+Mapping rules are attributes of the load balancer's SSL endpoint, which is identified by the certificate Common Name (CN).
+When you create mapping rules with:
+
+* Runtime Manager, you define mapping rules within a certificate definition. 
+* The CLI or CloudHub API, you must specify a certificate CN.
```
kbostrom: The style reference recommends against creating list items that end with a period and form a complete sentence only when appended to the list intro. Can you rework to comply?
See "bullet lists"

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
 
-Suppose you set up a DNS CNAME record from  `example.com` to `_lb-name_.lb.anypointdns.net`. You can then map `app.example.com` to a different deployed CloudHub Mule application name.
+=== Create Mapping Rules in Runtime Manager 
 
-For example, you can bind two hostnames for redirect:
+To create mapping rules in an SSL endpoint for a DLB in Runtime Manager:
+
+. From Anypoint Platform, select *Runtime Manager*.
+. Select *Load Balancers* in the left menu.
+. Click a load balancer name, then click a certificate name.
```
kbostrom:
```suggestion
. Click a load balancer name, and then click a certificate name.
```

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+. In the *URL Mapping Rules* section, click *Add New Rule*:
++
+.The screenshot shows (*1*) the *Add New Rule* icon and (*2*) the fields required to create a mapping rule.
+image::cloudhub-dlb-mapping-rules.png[the *Add New Rule* icon and the fields required to create a mapping rule]
+. Complete the fields:
++
+* *Input Path* (`inputUri`)
++ 
+The URI that the client requests: for example, `/{app}/`.
```
kbostrom: no period; GLOBAL for the first para following these field names
```suggestion
The URI that the client requests: for example, `/{app}/`
```

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
++
+The input URI is appended to the host header of the load balancer.
+* *Target App* (`appName`)
++
+The name of the CloudHub application that processes the request: for example, `{app}-example`.
```
kbostrom: ditto
```suggestion
The name of the CloudHub application that processes the request: for example, `{app}-example`
```

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+* *Output Path* (`appURI`)
++
+The URI string to pass to the app: for example, `/`.
```
kbostrom:
```suggestion
The URI string to pass to the app: for example, `/`
```

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
++
+The protocol on which the application listens:
++
+--
+** `http` (port 8091)
+** `https` (port 8092)
+** `ws` WebSockets (port 8091)
+** `wss` Web Service Security (8092)
+
+By default, the load balancer listens to external requests on HTTPS and communicates internally with your worker over HTTP. If you configured your Mule application within the VPC to listen on HTTPS, make sure you set *Protocol* to `https` when you create the mapping rule list.
```
kbostrom:
```suggestion
By default, the load balancer listens to external requests on HTTPS and communicates internally with your worker over HTTP. If you configured your Mule application within the VPC to listen on HTTPS, set *Protocol* to `https` when you create the mapping rule list.
```

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
-‘example.com/{mypattern}’ -> application: `app-{mypattern}` URI: /data
+anypoint.mulesoft.com/cloudhub/api/organizations/{orgid}/loadbalancers/{loadbalancerId}
 ----
 
-The example above causes that both `example.com/bookings` and `example.com/sales` match to `app-bookings/data` and `app-sales/data` respectively, as the variable `_mypattern_` holds these values.
-For `input=”bookings.example.com”`, the pattern can be resolved by assigning `_mypattern_=”bookings”` and for `input=`sales.example.com``, the pattern is resolved to assign `_mypattern_=”sales”`.
+See https://anypoint.mulesoft.com/exchange/portals/anypoint-platform/f1e97bc6-315a-4490-82a7-23abe036327a.anypoint-platform/cloudhub-api/version/V1/pages/home/[CloudHub API].
 
-Depending on your design, you can leverage your internal redirects to your endpoints by using either patterns or literal mappings.
+== Mapping Rule Priority
```
kbostrom: Leaving these questions here, because I had them before I read further. 
So...I wonder if this intro text is going to confuse the audience and perhaps we should get right to the point of naming the subcontent: 
What rule-processing "priority" mapping is vs literal mapping
How to change priority/mapping in RTM
How to change it in CLI
How to change in CloudHub API

********MY INITIAL MUSINGS*********
I keep reading over this section and something about it bothers me. 

This seems like it should be so simple, and yet "defining priority" sounds odd to me; isn't priority in a list obvious?

So...a set of rules is stored in a list (that is scanned by...?). Is this list a one-shot deal, or can rules be added to the end or injected anywhere within the list order later?

The higher that a rule is in that list, the more likely it will be used, because (the scan?) uses the first match that it comes across in the list, regardless of exactitude.

How do rules override each other, if the first one is chosen?

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
 
+Every rule must have a priority defined.
+The first rule in the list has a higher priority than rules lower in the list.
 
-== Creating Mapping Rules
+The DLB always uses the first rule that matches, regardless of whether a more exact match appears later in the list.
+When creating rules, pay attention to the order for each rule to avoid multiple rules overriding each other.
```
kbostrom: See above.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
-A mapping rule is a set of fields that define an input URL, and a set of fields that describe the output URL.
+In the following example, the `/api/v1/secondapp` URL maps to the `default-app` app:
 
-The input URL is described using a URI parameter that can be specified by the user. The `URI` is a string or pattern that describes the input URI.
+[%header,cols="10a,20a,20a,20a,10a"]
+|===
+|Priority|Input Path	|Target App| Output Path| Protocol
+|1 | /api/v1/ | default-app | /api/v1/ | http
+|2 | /api/v1/secondapp/ | second-app | /api/v1 |	http
+|===
```
kbostrom: spacing okay?

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+|Priority|Input Path	|Target App| Output Path| Protocol
+|1 | /api/v1/secondapp/ | second-app | /api/v1 |	http
+|2 | /api/v1/ | default-app | /api/v1/ | http
+|===
 
-* `appName`
-+
-Outputs the application name where the request is forwarded.
-* `appURI`
+=== Change the Priority of Mapping Rules in Runtime Manager
```
kbostrom: OH...this answers one of my questions....

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
-+
-Outputs the application name where the request is forwarded.
-* `appURI`
+=== Change the Priority of Mapping Rules in Runtime Manager 
+
+To change the priority of mapping rules in Runtime Manager:
+
+. From Anypoint Platform, select *Runtime Manager*.
+. Select *Load Balancers* in the left menu.
+. Click a load balancer name, then click a certificate name.
```
kbostrom:
```suggestion
. Click a load balancer name, and then click a certificate name.
```

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+. In the *URL Mapping Rules* section, mouse over the rule to display the handle:
```
kbostrom: "mouse over" > "hover over"

kbostrom: I suppose your audience is familiar with what a "handle" is or looks like? New to me.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
 +
-Specifies the URI string that is passed to the resolved application.
+.The arrow shows the handle to use to reorder a mapping rule.
+image::cloudhub-dlb-mapping-rules-reorder.png[Handle to use to reorder a mapping rule]
+. Grab the handle to move the rule up or down in the list.
```
kbostrom: grab = click and hold?

kbostrom: Maybe "list mapping" vs. "index mapping" vs. "patch mapping"?

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
 
-You can define both input and output URLs using patterns or literal strings.
+See xref:anypoint-platform-cli-commands.adoc#cloudhub-load-balancer-mappings-add[cloudhub load-balancer mappings add].
 
-Mapping rules are attributes of the load balancer's SSL endpoint, which is identified by the certificate name.
-When you create a mapping rule, you must specify a certificate CN. Omitting the `[certificateName]` parameter adds the mappings to the default endpoint.
+=== Change the Priority of Mapping Rules Using the CloudHub API 
 
-If your SSL endpoint sets a https://en.wikipedia.org/wiki/Wildcard_certificate[wildcard certificate], and you want to use the subdomain portion in a mapping rule, you can use the pre-defined `{subdomain}` variable.
+When using the API to create a mapping rule, you cannot specify a priority order.
```
kbostrom: So...how are the rules ordered when created?

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+However, you can update the order of the rules later by sending a `PATCH` request to the load balancer endpoint:
 
-The rule which is defined first has high priority against other ones defined after it. This means that the first matched rule will be applied.
-You can create, view and delete existing rules using the xref:anypoint-platform-cli.adoc#cloudhub-load-balancer-mappings-add[`mappings add`], xref:anypoint-platform-cli.adoc#cloudhub-load-balancer-mappings-describe[`mappings describe`] and xref:anypoint-platform-cli.adoc#cloudhub-load-balancer-mappings-remove[`mappings remove`] commands respectively.
+[source,text,linenums]
+----
+anypoint.mulesoft.com/cloudhub/api/organizations/{orgid}/loadbalancers/{loadbalancerId}
+----
 
+To determine the load balancer ID, perform a `GET` request to your endpoint:
```
kbostrom: According to our font use rules, HTTP requests do not get special treatment other than their common ALLCAP names.
```suggestion
To determine the load balancer ID, perform a GET request to your endpoint:
```

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+** `http` (port 8091)
+** `https` (port 8092)
+** `ws` WebSockets (port 8091)
+** `wss` Web Service Security (8092)
+
+By default, the load balancer listens to external requests on HTTPS and communicates internally with your worker over HTTP. If you configured your Mule application within the VPC to listen on HTTPS, make sure you set *Protocol* to `https` when you create the mapping rule list.
+--
+
+You can use either literal strings or patterns to define the input path. 
+For the output path, you can use only literal strings.
```
kbostrom: I kept rereading this topic set, and discovered that it would be more clear to me if there were a concept text about pattern mapping vs. literal mapping before getting into the priority texts.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+----
 
-The external load balancer domain name depends on the unique name you assign to it. In the following mapping rule examples, the load balancer is `lb-demo`.
 
-By default, your load balancer listens to external requests on HTTPS and communicates internally with your worker over HTTP. If you configured your Mule application within the VPC to listen on HTTPS, make sure you set `upstreamProtocol` to `HTTPS` when creating the mapping list using the xref:anypoint-platform-cli.adoc#cloudhub-load-balancer-mappings-add[`load-balancer mappings add`] command.
+== Using a Literal 1:1 Mapping
 
-=== URL Mapping
+Instead of using a pattern, you can map literal app names.
+In this case, the rules must match the applications exactly.
```
kbostrom: Is there more to say about this, such as relative pros and cons to the pattern vs. literal choice?
I guess this comes under the category of my need to understand the mapping methods earlier.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
 |===
-|Rule # |Input URL 2+^| Output URL
-|   |    *URI*   |       *appName*   |   *appURI*
-| 0 | /{app}/    | {app}             | /
+| Rule # | *Input Path* | *Target App* | *Output Path* | *Protocol*
+| 1 | / | myApp | / | http 
 |===
 
-This rule maps `lb-demo.lb.anypointdns.net/{app}` to `{app}.cloudhub.io`. `{app}` is a pattern for any application name you pass in the inbound request.
+This rule matches the `dev-misc-myhellotest-v1-myapp` app if the input URI is `/misc/myhellotest/v1/ping`, but not `/misc/myhellootest/v1/ping`:
```
kbostrom: no comma, which adds a potential other meaning:
```suggestion
This rule matches the `dev-misc-myhellotest-v1-myapp` app if the input URI is `/misc/myhellotest/v1/ping` but not `/misc/myhellootest/v1/ping`:
```

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+| 2 | /v1/applicationY | application-y | /api/ | http
+| 3 | /v2/applicationx | application-x-v2 | /api/ | http
 |===
-This rule automatically maps any request passed to a subdomain of `example.com` to the corresponding `appName`:
 
-* Passing `api.example.com` would redirect to `api.cloudhub.io` +
-* Passing `application.example.com` is mapped to `application.cloudhub.io`.
 
-The same applies to the https://en.wikipedia.org/wiki/Subject_Alternative_Name[Subject Alternative Names] (SANs) of your SSL endpoints.
+== Using Patterns in Mapping Rules
```
kbostrom, line 108: Shouldn't this topic come earlier in the set?
See earlier mention of my need to understand pattern vs. literal earlier.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+
+A _pattern_ is a string that defines a template for matching an input text.
+The value in curly brackets (`{   }`) is treated as a variable.
```
kbostrom:
```suggestion
The value in curly brackets (`{   }`) is treated as a variable:
```

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
 
-* `dev.example.com` redirects to `dev-app.cloudhub.io`.
-* `qa.example.com` redirects to `qa-app.cloudhub.io`.
+[%header,cols="10a,20a,20a,20a,10a"]
+|===
+| Rule # | *Input Path* | *Target App* | *Output Path* | *Protocol*
+|1 | example.com/{myPattern} | app-{myPattern} | /data | http
+|===
 
+In this case, both `example.com/bookings` and `example.com/sales` match to `app-bookings/data` and `app-sales/data` respectively, because the variable `_myPattern_` holds these values.
```
kbostrom:
```suggestion
In this case, both `example.com/bookings` and `example.com/sales` match `app-bookings/data` and `app-sales/data` respectively, because the variable `_myPattern_` holds these values.
```
Had you meant to use "map to" instead of "match to"?

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
 
-=== 1:1 Mapping
+If the input is:
 
-If you have only one application, you can map the literal app name.
+* `bookings.example.com`, the pattern resolves if the value of `_myPattern_` is `bookings`. 
+* `sales.example.com`, the pattern resolves if the value of `_myPattern_` is `sales`.
```
kbostrom: This is another situation in which the list items grammatically complete the intro snippet and so should not end with periods (as independent list items).
Suggest rewording to avoid this grammatical structure GLOBALLY.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+| 1      | /            | {subdomain}  | /             | https
+|===
+
+This rule automatically maps any request passed to a subdomain of `example.com` to the corresponding `appName`:
+
+* Passing `api.example.com` redirects to `api.cloudhub.io`.
+* Passing `application.example.com` maps to `application.cloudhub.io`.
 
-=== Ordering and Prioritizing Rules
+The same behavior applies to the https://en.wikipedia.org/wiki/Subject_Alternative_Name[Subject Alternative Names] (SANs) of your SSL endpoints.
```
kbostrom: I suggest rewording to make clear to what "the same behavior" applies, to make this subtopic independent and therefore reusable.

modules/ROOT/pages/lb-mapping-rules.adoc
```diff
@@ -8,165 +8,292 @@ image:logo-hybrid-disabled.png[link="/runtime-manager/deployment-strategies", ti
...
+
+Instead of using a wildcard certificate, which permits any subdomain application name to pass through the DLB, you can specify which subdomain names to allow.
+To do this, you must have different SANs configured for a certificate's common name.
```
kbostrom: "different" from...? One another?

kbostrom (approved):
Done. See comment in Slack completion note.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: Yes.

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: There's a picture...?

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: The problem is...no one understands mapping rules...

modules/ROOT/pages/lb-mapping-rules.adoc
hannanelson: I will address this in the future.

PR #256, opened 2019-12-19
Update extending-the-runtime-manager-agent.adoc

amrsalah-acc:
External message handers to be External message handlers


hannanelson (approved):
LGTM!

PR #261, opened 2019-12-31
Create vpn-update-arm.adoc

sjsheridan:
Adding a new section for updating VPN connections. This was not previously documented, but needs defining.
This should be added between the Create and Troubleshoot sections.
Please add @danieldiaz-mule for review.


sjsheridan (commented):
Under Static Routes, we need to make a small change. 
6. Click Apply Changes to add the new rules.
7. (as it is)
8. Update the configuration on your VPN endpoint.
See the documentation for your device.
When the tunnel is active, the route table updates automatically.

Under Dynamic Routes, we should switch the order of point 1 and 2, to make it consistent with the static section.

The term "route table" should not be hyphenated. 

Then everything else as it is.

danieldiaz-mule (approved):
LGTM

hannanelson (approved):

PR #268, opened 2020-01-16
added "region" to all URLs, including the graphic.

hannanelson:


jorgegarciamule (approved):
Looks good! Thanks

modules/ROOT/pages/cloudhub-networking-guide.adoc
```diff
@@ -10,15 +10,16 @@ This guide is for cloud deployments using the cloud-based version of Runtime Man
 
 CloudHub provides a variety of tools to architect your integrations and APIs so they are maintainable, secure, and scalable. This guide covers the basic network architecture, DNS, and firewall rules.
 
-image::cloudhub-networking-guide-1.jpg[CloudHub+Networking+Guide-1]
+.The image shows the CloudHub Network Overview.
+image::ch-networking-guide.png[CloudHub Network Overview]
```
rddheath, line 14: Questions about the diagram:

- Why do the lines that flow either from a node to text or text to node (I'm not sure which) have the same appearance as the lines between nodes?
- Why is there a stronger shadow on the load balancer than on the worker nodes? Why are you using shadows? They don't seem necessary.
- What is the relationship of the text next to the load-balancer node to that node?
- Why do the two lines extending from the worker nodes converge on empty space?
- If the node labeled "Optional VPC" is optional, where are the workers if that VPC is not used?
- What is the relationship between the load-balancer node and the worker nodes?
- In which direction is data flowing between the load-balancer nodes and the worker nodes?
- What is the relationship of the text labels about the external and internal DNSs to the worker nodes?
- What is the relationship of the external VPC network to the rest of the diagram?
- Why is there no text after the diagram to explain what it depicts, since what it depicts is not at all obvious?

Question about the caption: Is the diagram an overview of the entire CloudHub network? Or does it show only a portion of it, and that portion related somehow to two apps? The caption doesn't help me understand what the diagram depicts. There isn't any "network overview" in CloudHub, as far as I know.

modules/ROOT/pages/cloudhub-networking-guide.adoc
```diff
@@ -10,15 +10,16 @@ This guide is for cloud deployments using the cloud-based version of Runtime Man
...
 
 == Load Balancing
 
 CloudHub provides a load balancing service for all integrations. You can use the default load balancing service or obtain a xref:cloudhub-dedicated-load-balancer.adoc[dedicated load balancer].
 
 The CloudHub load balancing service does round robin load distribution across workers, which allows workers to scale linearly as they receive more requests. The load balancing service also provides transparent switchover when an application is upgraded. See xref:managing-applications-on-cloudhub.adoc[Managing Applications on CloudHub].
 
-Each application deployed on CloudHub has a `CNAME` record that refers to the load balancer, for example `myapp.cloudhub.io`. Mule apps deployed in CloudHub must listen on host `0.0.0.0` and to either the HTTP port (`8081`) or to the HTTPS port (`8082`).
+Each application deployed on CloudHub has a `CNAME` record that refers to the load balancer, for example `myapp.region.cloudhub.io`. Mule apps deployed in CloudHub must listen on host `0.0.0.0` and to either the HTTP port (`8081`) or to the HTTPS port (`8082`).
```
rddheath, line 26: "Each application deployed on CloudHub has a `CNAME` record that refers to the load balancer, for example `myapp.region.cloudhub.io`." The use of "for example" like this make the sentence a run-on sentence that should be split into two sentences.

modules/ROOT/pages/cloudhub-networking-guide.adoc
```diff
@@ -45,7 +46,7 @@ If other values for `http.port` and `https.port` are specified in the `mule-app.
 
 [NOTE]
 --
-For each request a client makes through CloudHub's load balancer (`myapp.cloudhub.io`), the load balancer maintains two connections. One connection ​is from​ the client and one is to your worker.  For each connection, the load balancer manages an idle timeout of 300 seconds that is triggered when no data is sent over ​either​ connection. If no data is sent or received during this time period, the load balancer closes ​both connections.
+For each request a client makes through CloudHub's load balancer (`myapp.region.cloudhub.io`), the load balancer maintains two connections. One connection ​is from​ the client and one is to your worker.  For each connection, the load balancer manages an idle timeout of 300 seconds that is triggered when no data is sent over ​either​ connection. If no data is sent or received during this time period, the load balancer closes ​both connections.
```
rddheath, line 53: About "and one is to your worker": shouldn't "worker" be plural? After all, a load balancer balances requests among more than one worker, right?
About "the load balancer manages an idle timeout of 300 seconds that is triggered...": The timeout is "idle"? And does the load balancer "manage an idle timeout...that is triggered", or does it simply close connections when they have been idle for 300 seconds?

modules/ROOT/pages/cloudhub-networking-guide.adoc
```diff
@@ -57,17 +58,17 @@ The following DNS records for your applications are exposed when they are deploy
...
 |===
 
 In certain situations, you may want to know an application’s internal IP addresses, for example:
 
 * When communicating with the worker directly within the customer's xref:virtual-private-cloud.adoc[Anypoint VPC] without sending data over the public internet
 * When setting up your own load balancer
 
-You can access these IP addresses through the `mule-worker-internal-myapp.cloudhub.io` record from within the Anypoint VPC. If you access the workers directly, any load distribution benefits from the CloudHub load balancing layer are lost.
+You can access these IP addresses through the `mule-worker-internal-myapp.region.cloudhub.io` record from within the Anypoint VPC. If you access the workers directly, any load distribution benefits from the CloudHub load balancing layer are lost.
```
rddheath, line 75: About "If you access the workers directly, any load distribution benefits from the CloudHub load balancing layer are lost.": are you saying that, if I access workers directly, I won't benefit from the capabilities of the load balancer to balance load across the workers? If so, I didn't understand that without reading the sentence a few times. For example, if I don't benefit from something, how can I "lose" the benefits which I never had? Also, "load distribution benefits" and "CloudHub load balancing layer" are two noun strings that make the sentence hard to parse.

rddheath (changes requested):

rddheath (approved):

PR #273, opened 2020-01-22
Docs 6108 provide some examples hn

hannanelson:
Clarified that these ARE examples, edited topic to adhere to style guidelines.


mcocaro (approved):
LGTM

modules/ROOT/pages/about-logging-of-payload-formats.adoc
```diff
@@ -1,27 +1,30 @@
...
+image:logo-rtf-disabled.png[link="/runtime-manager/deployment-strategies#anypoint-runtime-fabric", title="Runtime Fabric"]
 
-Some payload formats can't be converted to plain text. Because of this, the Runtime Manager Agent differentiates formats accordingly and only converts those that are considered to work.
 
-When the payload format isn't of a convertible type, the agent only passes on the full qualified name of the payload class type, not its content.
+To send message payload data to external analytics software, Runtime Manager agent must first serialize and convert the data to text.
 
-[WARNING]
-The information on this document only applies as of version xref:release-notes::runtime-manager-agent/runtime-manager-agent-1.6.3-release-notes.adoc[1.6.3] of the Runtime Manager agent. Prior versions attempt to convert all formats, causing failures on some.
+Because some payload formats can't be converted to plain text, Runtime Manager agent only converts those formats that are considered to work.
```
kbostrom: That darn "only"!
```suggestion
Because some payload formats can't be converted to plain text, Runtime Manager agent converts only those formats that are considered to work. 
```

modules/ROOT/pages/about-logging-of-payload-formats.adoc
```diff
@@ -1,27 +1,30 @@
...
 
-Some payload formats can't be converted to plain text. Because of this, the Runtime Manager Agent differentiates formats accordingly and only converts those that are considered to work.
 
-When the payload format isn't of a convertible type, the agent only passes on the full qualified name of the payload class type, not its content.
+To send message payload data to external analytics software, Runtime Manager agent must first serialize and convert the data to text.
 
-[WARNING]
-The information on this document only applies as of version xref:release-notes::runtime-manager-agent/runtime-manager-agent-1.6.3-release-notes.adoc[1.6.3] of the Runtime Manager agent. Prior versions attempt to convert all formats, causing failures on some.
+Because some payload formats can't be converted to plain text, Runtime Manager agent only converts those formats that are considered to work. 
+When the payload format can't be converted, the agent passes only the fully qualified name of the payload class type, not the payload message content.
```
kbostrom, line 18: Yay only!

modules/ROOT/pages/about-logging-of-payload-formats.adoc
```diff
@@ -1,27 +1,30 @@
...
+
+[NOTE]
+Runtime Manager agent version 1.6.2 and earlier attempts to convert all payload formats, which results in failures for some formats.
 
 == Default Conversion Behavior
 
-The Runtime Manager agent only passes on payload messages where the *Converted to String* column reads `TRUE` in the table below:
+The following table lists the payload formats and whether Runtime Manager agent converts them to String and passes the payload message.
```
kbostrom:
```suggestion
The following table lists the payload formats and specifies whether Runtime Manager agent converts them to the String type and passes the payload message.
```

modules/ROOT/pages/about-logging-of-payload-formats.adoc
```diff
@@ -1,27 +1,30 @@
...
+If the *Converted to String* column shows `FALSE`, Runtime Manager agent passes only the payload class type.
```
kbostrom, line 26: This is a deeply debated issue on which I would like your opinion: why do we use mono/syntax font when documenting a GUI? In this sentence, it seems to me that FALSE is better than `FALSE`.
I would like to limit ` ` to literal user syntax input and system syntax output, just as I would like to limit * * to the labels of items actually clicked or selected. I think we have way more formatting than we need, and that it distracts from narrative flow. But there might be legacy reasons for this of which I am not yet aware, or other opinions that I have not yet understood....
Please don't DO anything; I am just curious about your take on these issues.

modules/ROOT/pages/about-logging-of-payload-formats.adoc
```diff
@@ -85,13 +88,17 @@ The Runtime Manager agent only passes on payload messages where the *Converted t
 |	|Multipart|	"multipart/mixed, multipart/alternative, multipart/digest multipart/parallel"	|FALSE||
 |===
 
-== Customizing Conversion Behavior
+== Customize Conversion Behavior
```
kbostrom, line 91: Here's another issue that I would like input on: imperative vs. gerund in titles.
I've accepted that this department uses the imperative for tasks.
But topics like this one have no steps...which is how I define a task.
I would use Customizing or Customization here...but please don't DO anything but comment, if you have time.

modules/ROOT/pages/about-logging-of-payload-formats.adoc
```diff
@@ -85,13 +88,17 @@ The Runtime Manager agent only passes on payload messages where the *Converted t
...
+
+You can override the default behavior and configure the agent to forward all payload formats or none by adding
+properties to the `conf/mule-agent.yml` file.
```
kbostrom, line 94: ditto here: it's a file name, distinct from the narrative by its very format. Why format it as if it were syntax?
OK...I'll stop....

modules/ROOT/pages/about-logging-of-payload-formats.adoc
```diff
@@ -112,14 +120,15 @@ You may want to force the exclusion of certain formats from the external logs. T
...
 
-You may also want to force the conversion and forwarding of formats that are considered to not be convertible.
+=== Include Payloads 
 
-[WARNING]
-Note that by doing this there's a risk of breaking the payload.
+You might want to convert and forward payload formats that Runtime Manager agent considers not convertible.
 
+[IMPORTANT]
+If you force conversion for all payload formats, the payload might break.
```
kbostrom, line 128: What does it mean to have a payload "break"?

kbostrom: Also...are the steps that follow doing this: forcing the conversion of all formats?

kbostrom (approved):
Here you go!

modules/ROOT/pages/about-logging-of-payload-formats.adoc
hannanelson: In this case, the `FALSE` refers to the value in the table (not GUI). Technically, this one should probably be regular font.

As for bold for GUI items, I have been trained to tag actual GUI text - for translation purposes. I know we don't have semantic markup, but this is a hard habit to break.

Because the UX style is Sentence case (although not really consistently followed), I believe the bold helps with scanning and comprehension because it distinguishes the UI element in the middle of a sentence. 

Finally, I think we need to simplify our style guidelines. The fewer things writers need to keep in our heads, the easier it is to follow the style consistently (and ultimately less work for you and peer reviewers). So, for example, a guideline that's "bold all UI elements" is much easier to remember than "bold UI elements only when instructing the user to click or select". Another example of style simplification would be "Always include terminating punctuation in list items". That eliminates the whole "is this a complete sentence or not?" and "What do I do if I have a complete sentence after the sentence fragment?" lookups (or not) in the SG.

My (lofty) goal is to have the style in my head, so I don't have to look stuff up.

modules/ROOT/pages/about-logging-of-payload-formats.adoc
hannanelson: Although this topic has no steps, it does list tasks (do this to accomplish that). One-step tasks are still tasks, but don't have "steps".

I prefer imperative for tasks - and I believe this is how the industry is moving.

In this case, the issue was the reader couldn't see that there were examples. I wanted to emphasize the taskiness. Perhaps wrongly.

modules/ROOT/pages/about-logging-of-payload-formats.adoc
hannanelson: The formatting guidelines when I first joined were to format filenames (and URLs) as code - there was a proposal that Mysti put together. I can't find it now and I'm not sure it got migrated.

I feel less strongly about this font change, but some filenames are not as obvious, so...

modules/ROOT/pages/about-logging-of-payload-formats.adoc
hannanelson: I fixed the FALSE. I supposed in the olden days, this would be "FALSE".

modules/ROOT/pages/about-logging-of-payload-formats.adoc
kbostrom: Gotcha. I didn't even consider "hidden" single-step instructions. My bad. Thanks!

modules/ROOT/pages/about-logging-of-payload-formats.adoc
kbostrom: Yeah, I hear you about making the style rules simpler than English grammar and punctuation, with all its deviations and exceptions. Thanks for reminding me that my second most important goal is making writers' lives easier....
I'm really sorry that UX decided on sentence-style labeling in the UI. It might look better there, but it introduces this very problem for docs. I conversed a bit with Tony about it, who agrees but indicated that consensus made this a done deal.

kbostrom (approved):
Just conversation.

PR #282, opened 2020-02-06
Update cloudhub-networking-guide.adoc

jmjdamen:
url fix

### DIFF
--- a/modules/ROOT/pages/cloudhub-networking-guide.adoc
+++ b/modules/ROOT/pages/cloudhub-networking-guide.adoc
@@ -133,7 +133,7 @@ supports some of the same regions that the U.S. control plane supports, the DNS 
 are different. For more on the EU control plane, see
 xref:eu-control-plane::index.adoc[About the EU Control Plane].
 
-Deploying to a region also affects your internal and external worker DNS address. For example, if you deploy in Ireland in the US control plane, the DNS records for internal and external IP addresses are `mule-worker-myapp.ir-e1.cloudhub.io` and `mule-worker-internal-myapp.ir-e1.cloudhub.io`.
+Deploying to a region also affects your internal and external worker DNS address. For example, if you are using the US control plan and deploy to the Ireland region, the DNS records for internal and external IP addresses are `mule-worker-myapp.ir-e1.cloudhub.io` and `mule-worker-internal-myapp.ir-e1.cloudhub.io`.
 
 == Firewall Rules and Port Access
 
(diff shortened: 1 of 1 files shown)


hannanelson:
@sjsheridan - please confirm.

sjsheridan (changes requested):
This change is incorrect. The change would be right for the EU control plane, but that is not covered in this section. The original URLs are correct for the US control plane, which is the example being given here.

Suggest calling out that this is for US instead. I.e.

> Deploying to a region also affects your internal and external worker DNS address. For example, if you deploy in Ireland in the US control plane, the DNS records for internal and external IP addresses are `mule-worker-myapp.ir-e1.cloudhub.io` and `mule-worker-internal-myapp.ir-e1.cloudhub.io`.

sjsheridan (approved):

fermujica (approved):

modules/ROOT/pages/cloudhub-networking-guide.adoc
```diff
@@ -133,7 +133,7 @@ supports some of the same regions that the U.S. control plane supports, the DNS
 are different. For more on the EU control plane, see
 xref:eu-control-plane::index.adoc[About the EU Control Plane].
 
-Deploying to a region also affects your internal and external worker DNS address. For example, if you deploy in Ireland, the DNS records for internal and external IP addresses are `mule-worker-myapp.ir-e1.cloudhub.io` and `mule-worker-internal-myapp.ir-e1.cloudhub.io`.
+Deploying to a region also affects your internal and external worker DNS address. For example, if you deploy in Ireland in the US control plane, the DNS records for internal and external IP addresses are `mule-worker-myapp.ir-e1.cloudhub.io` and `mule-worker-internal-myapp.ir-e1.cloudhub.io`.
```
Cristian-Venticinque: if you deploy **in** Ireland **in** the US control plane > This sentence is creating a bit of confusion for me, for example the first part means the user deploys to Ireland or the user is based in Ireland while doing the deployment? 

If possible, I'd prefer to replace "in" with another word that specifically describes the scenario, for example a suggestion could be (assuming what I say is technically correct):

"For example, if you deploy **to** Ireland **using** the US control plane..."

Cristian-Venticinque (approved):
Please check comment before merging

modules/ROOT/pages/cloudhub-networking-guide.adoc
sjsheridan: "if you are using the US control plane, and deploy to the Ireland region..."

PR #286, opened 2020-02-11
[ION-7231] Deprecate metadata and replay feature for free and trial users (19-Feb)

kpurvesh:


kpurvesh:
This needs to be merged on 19th Feb after the cloudhub release.

hannanelson:
LGTM.
@kpurvesh - It sounds like there will be a UI change, too? If so, I'll need to update some screenshots.
And, we'll also need Release Notes.

hannanelson (approved):

PR #304, opened 2020-03-04
Added comments from ticket 6753.

hannanelson:


evinaku (commented):
swap instantiates with resumes communications with the queue

evinaku (approved):

modules/ROOT/pages/cloudhub-hadr.adoc
```diff
@@ -13,7 +13,9 @@ image::hadr-aws-global-infrastructure.png[]
 For example, if the US East region is unavailable, the CloudHub management UI, as well as the various REST services that enable deployments, are unavailable until the region's availability is restored.
 New applications can't be deployed while US East is down.
 
-CloudHub provides an internal messaging mechanism, in the form of persistent queues, that is used for message reliability. The persistent queues are highly available within a region. However, these persistent queues are lost when the region is unavailable (usually a few seconds or minutes), which could result in some data loss.
+CloudHub provides an internal messaging mechanism, in the form of persistent queues, that is used for message reliability.
+While persistent queues are highly available within a region, they might not accessible if the region or part of the region is unavailable (usually a few seconds or minutes), which could result in some data loss.
```
wwarner-salesforce:
```suggestion
While persistent queues are highly available within a region, they might not be accessible if the region or part of the region is unavailable (usually a few seconds or minutes), which could result in some data loss.
```

wwarner-salesforce (approved):

PR #307, opened 2020-03-10
Update continuous-deployment.adoc

askamycki:
The link for Runtime Manager Registry API is broken (deprecated maybe?)


hannanelson (approved):
Thanks for finding/fixing!

PR #313, opened 2020-03-18
Fixed ports in diagram.

hannanelson:


DavoCoder:
Looks good! thanks for the fix

sjsheridan (changes requested):
This change is not needed - the diagram is already correct. Port 8081/82 are used here because there is no DLB. The next diagram includes the DLB and uses 8091/92.

PR #317, opened 2020-03-30
Update limit on logs API

sclowes19:
Please see https://wiki.corp.mulesoft.com/x/r5o5BQ


modules/ROOT/pages/list-all-logs.adoc
```diff
@@ -13,7 +13,7 @@ image:logo-rtf-disabled.png[link="/runtime-manager/deployment-strategies", title
 [TIP]
 This document deals with calling the xref:cloudhub-api.adoc[CloudHub API]. For instructions on how to perform these same tasks more easily via the Runtime Manager UI, see xref:cloudhub.adoc[CloudHub] and its child pages.
 
-Use the GET operation on `/applications/{application}/logs` resource to list all log messages. This call returns a maximum of 10,000 lines of log data for each invocation. However, using repeated calls with contiguous time windows as search criteria, the entire log may be downloaded.
+Use the GET operation on `/applications/{application}/logs` resource to list all log messages. This call returns a maximum of 2,500 lines of log data for each invocation. However, using repeated calls with contiguous time windows as search criteria, the entire log may be downloaded.
```
alexsmirnov: 1. This is POST request.
2. This is log search api, and not supposed to list all logs ( while it's possible, request will most likely timed out because of amount of records involved ).
3. Production limit is 2000
4. Default limit is 100
5. Output JSON example is wrong
6. This api does not exist

sclowes19: @alexsmirnov I'm not sure what to say to that, the customer i using it and is complaining that the limit was reduced and we did not document it. Please see https://na101.salesforce.com/5002T00001A3Jv6 and https://wiki.corp.mulesoft.com/x/r5o5BQ

Would you propose we just delete the page? Happy to do that if this is easier, I'm simply trying to make the situation incrementally better, please help me prevent this data staying wrong

hannanelson: I've removed the page - I will merge in the morning. Speak now or forever hold your peace.

hannanelson (approved):

PR #320, opened 2020-04-02
Added note per bug

hannanelson:
Assuming bug is correct.

### DIFF
--- a/modules/ROOT/pages/anypoint-platform-cli-commands.adoc
+++ b/modules/ROOT/pages/anypoint-platform-cli-commands.adoc
@@ -487,7 +487,8 @@ This command does not take any options, except for the default ones: `--help`, `
 > api-mgr api delete [options] <apiInstanceId>
 ----
 
-This command deltes the API Instance passed in `<apiInstanceId>`. +
+This command deletes the API instance passed in `<apiInstanceId>`.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/anypoint-platform-cli-commands.adoc
```diff
@@ -496,7 +497,10 @@ This command does not take any options, except for the default ones: `--help`, `
 > api-mgr api deploy [options] <apiInstanceId>
 ----
 
-This command deploys the API instanced passed in <apiInstanceId> to the deployment target specified using the options described below.
+This command deploys the API instance passed in <apiInstanceId> to the deployment target specified using the options described below.
+
+[NOTE]
+This command is only supported for non-proxy API apps.
```
sbehrends: This command only applies to APIs Managed that have "Endpoint with Proxy" selected or if they where created using `--withProxy` via the CLI, so the proper wording I suggest is:

```suggestion
This command is only supported for endpoints with proxy.
```

sbehrends (changes requested):

sbehrends (approved):
LGTM 🚀

Mulesoft-Mysti (approved):

PR #323, opened 2020-04-16
Update managing-cloudhub-specific-settings.adoc

pcadoppi:
Support can’t help with this and will redirect to Account representative. CSM is better alternative.


hannanelson (approved):
LGTM

PR #329, opened 2020-04-22
Added --region, updated example ports

hannanelson:
Ran the agent command with --help and compared with what was documented. Only --region was missing.

### DIFF
--- a/modules/ROOT/pages/installing-and-configuring-runtime-manager-agent.adoc
+++ b/modules/ROOT/pages/installing-and-configuring-runtime-manager-agent.adoc
@@ -249,6 +249,14 @@ This name must be unique within the business group's environment.
 |Specifies the proxy configuration to use when registering with the connection.
 
 See xref:rtm-agent-proxy-config.adoc[Connect the Agent Through a Proxy Server].
+
+
+|`-R`
+
+`--region`
+|`_region_`
+|Specifies the region to deploy to. Available regions are `prod` and `eu1`.
+
 |===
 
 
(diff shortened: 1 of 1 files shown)


bauerchristina (approved):

modules/ROOT/pages/installing-and-configuring-runtime-manager-agent.adoc
```diff
@@ -378,22 +386,22 @@ Do not use these parameters with the `-I` or `S` parameters to configure REST AP
 `--amc-host`
 | `_amc-host_`
 
-|Specifies the service location URL of your local instance of Runtime Manager, such as `+https://10.0.0.1:8080/hybrid/v1+`.
+|Specifies the service location URL of your local instance of Runtime Manager, such as `+https://10.0.0.1:443/hybrid/v1+`.
```
rddheath: I notice that your examples are introduced differently:

- This one with a comma and "such as"
- The next with a colon and "for example,"
- The next two with a comma and "for example,"

What I would do is end with a period each sentence that begins with "Specifies". Then, use this construction: "Example: `+https...whatever`", with no period at the end.

Whether you do that or not, each example should be introduced in the same way.

modules/ROOT/pages/installing-and-configuring-runtime-manager-agent.adoc
```diff
@@ -378,22 +386,22 @@ Do not use these parameters with the `-I` or `S` parameters to configure REST AP
...
+|Specifies the service location URL of your local instance of MCM: for example, `wss://10.0.0.2:8889/mule`.
 
 You can test whether the service is available at `_MCM_HOST_/mule`.
 
 |`-C`
 
 `--cs-host`
 |`_core-services-host_`
-|Specifies the service URL of your local instance of Access Management, for example, `+https://10.0.0.3:8080/accounts+`.
+|Specifies the service URL of your local instance of Access Management, for example, `+https://10.0.0.3:443/accounts+`.
```
rddheath: Should "the service URL" be "service location URL", as is the case at lines 389, 396, and 416?

rddheath (approved):

modules/ROOT/pages/installing-and-configuring-runtime-manager-agent.adoc
hannanelson: This is different than the SG guidance, but made them consistent.

PR #337, opened 2020-05-01
AZ distance update

jmjdamen:
AZ distance is not aligned with the AWS docs.
https://aws.amazon.com/about-aws/global-infrastructure/regions_az/

"AZ’s are physically separated by a meaningful distance, many kilometers, from any other AZ, although all are within 100 km (60 miles) of each other."


Terry-He:
I don't think we should change this - the current description is not wrong - it is within 350mile of distances. And we should not quote AWS - that could change at anytime in the future if we are brining in new AZs or AWS is opening new regions.

hannanelson:
Closing per Terry's comment.

PR #339, opened 2020-05-01
Moved topic closer to deploying per suggestion.

hannanelson:


modules/ROOT/nav.adoc
```diff
@@ -4,6 +4,7 @@
 * xref:deployment-strategies.adoc[Deployment Options]
  ** xref:deploying-to-cloudhub.adoc[Deploying to CloudHub]
  ** xref:deploying-to-your-own-servers.adoc[Deploying to Your Servers]
+  *** xref:rtm-agent-whitelists.adoc[Ports, IP Addresses, and Hostnames to Whitelist]
```
Cristian-Venticinque, line 7: Is there any chance to move the info from this article to the parent article as a new section? 
The only reason I'm suggesting that is per our guideline to avoid single child topics, however if the topic has enough content to be in its own article and cannot be merged, then its ok.

Cristian-Venticinque (approved):

modules/ROOT/nav.adoc
hannanelson: I think it needs to stand alone, but point taken about single child topics. I did think about that - we got specific feedback that this is primarily related to deploying, secondarily to agent.

PR #340, opened 2020-05-04
Add a warning when retaining mule-agent.yml file

rodrigo-neira:
Many customers opt to retain their mule-agent.yml so they do not lose configurations already made.
But when upgrading to agent versions >1.12.0 or >2.2.0 there are some required changes in the endpoints. If not making these url changes, it will generate connection issues. 
Please add a note stating that if retaining the file ensure to have latest required endpoints.


mcocaro (approved):
LGTM! Great recommendation Thank you!

hannanelson (approved):

PR #342, opened 2020-05-06
Clarified restart is for unicast only.

hannanelson:


bauerchristina (approved):
Thanks again! Now it is very clear, I think.

wwarner-salesforce (approved):

PR #363, opened 2020-06-15
Cloudhub pattern needs to be MM-dd HH:mm:ss.SSS to let Log-Entries ap…

ca-stefan-cordes:
…pear

With MM-dd HH:mm:ss nothing comes out of the API where you can download the log-files.
With MM-dd HH:mm:ss.SSS the log-entries appear as expected.

### DIFF
--- a/modules/ROOT/pages/custom-log-appender.adoc
+++ b/modules/ROOT/pages/custom-log-appender.adoc
@@ -85,7 +85,7 @@ Sends log data to the filesystem of the VM.
                 serverHeartbeatSendIntervalMs="${sys:logging.serverHeartbeatSendIntervalMs}"
                 statisticsPrintIntervalMs="${sys:logging.statisticsPrintIntervalMs}">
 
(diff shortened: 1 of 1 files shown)


MichelleZMY:
Hi thanks for your contribution. We have tested through those two Patterns it turned out both (with or without ".SSS") worked without an issue. If you are still seeing an issue with it, feel free to raise a case or discuss with us in the forum.

hannanelson:
Closed per dev.

ca-stefan-cordes:
Looks like the problem is when grabbing the logfiles with https://github.com/yevmel/logstash-input-cloudhub
maybe in conjunction with https://github.com/rails/rails/issues/22171
At least it is confusing when specify the 
 <PatternLayout pattern="[%d{MM-dd HH:mm:ss}] %-5p %c{0} [%t]: %m%n"/>
but log shows
11:40:16.722     06/23/2020 
image

At least it is fine for us and documentation can be stay as it is.

PR #364, opened 2020-06-15
Update vpn-about.adoc

sjsheridan:
DPD requirement was in the wrong place. Moved under static requirements as it should be configured on all types of VPN.


hannanelson (approved):
LGTM

PR #365, opened 2020-06-22
Add table on worker size section

smontimule:
Currently, in this page, under the "Worker Size" section, it's not indicated the storage capacity of each worker size, as opposed to https://docs.mulesoft.com/runtime-manager/deploying-to-cloudhub#worker-size-and-vcores
For that reason, I'm placing here the same table that's indicated in the previous link so that it will be the same.


hannanelson (approved):
LGTM

PR #375, opened 2020-07-13
Added per Christina

hannanelson:


bauerchristina:
Thank you for the update, @hannanelson. The content is clear now to me. 

I only think we should make the formatting for lb-name consistent. In "Additionally, a CloudHub dedicated load balancer has an internal domain name to be used by applications and clients within the Anypoint VPC. The structure is internal-<lb-name>.lb.anypointdns.net" lb-name is put in brackets whereas in the other places it is put in italics.

bauerchristina (approved):
Thanks again, Hanna.

wwarner-salesforce (approved):

PR #394, opened 2020-09-02
Update lb-mapping-rules.adoc

davidvancina-mulesoft:
I think this language clarifies a bit. Submitting for consideration...


hannanelson:
LGTM but I'd like @danieldiaz-mule to validate!

hannanelson (approved):

PR #395, opened 2020-09-09
Update Gov Cloud info

paigehargrave:


rishath-rias (commented):
Are we bundling it under US Control Plane table? Maybe we should have a separate table for GovCloud? Also for region maybe we say something like "US Gov West"?

Also, the DNS record should be in same format as other 'myapp.usg-w1.gov.cloudhub.io'

modules/ROOT/pages/cloudhub-networking-guide.adoc
```diff
@@ -124,6 +124,17 @@ covers the following regions.
 
 |===
 
+https://anypoint.mulesoft.com[Anypoint Platform] for the MuleSoft Government Cloud
```
priyaayyar: covers the following regions:

priyaayyar (dismissed):

modules/ROOT/pages/cloudhub-networking-guide.adoc
hannanelson: remove the "the"? I don't think we use articles with product names.

hannanelson: Agree with Priya.

paigehargrave: Removed the "the". Left the "." to be consistent with the other sections and minimize changes to your file.

paigehargrave: Left the "." to be consistent with other sections to minimize changes since I don't own this file.

hannanelson (approved):
LGTM

PR #402, opened 2020-10-12
Docs 8913 confusing runtime manager alerts placeholders hn

hannanelson:
[26-after-Alerts _ MuleSoft Documentation.pdf](https://github.com/mulesoft/docs-runtime-manager/files/5409857/26-after-Alerts._.MuleSoft.Documentation.pdf)

### DIFF
--- a/modules/ROOT/pages/alerts-on-runtime-manager.adoc
+++ b/modules/ROOT/pages/alerts-on-runtime-manager.adoc
@@ -379,68 +379,69 @@ The following table lists the variables that you can use with each alert.
 [%header,cols="2,30,60"]
 |===
 ||Alert |Variables
-| *All Applications*
-
-3+|*CloudHub Applications*
-.9+||*CPU usage - CloudHub* |`${severity}` `${state}` `${resource}` `${operator}` `${value}`
-| *Custom application notification* a|`${severity}` `${resource}` `${user}` `${priority}` `${transactionId}` `${message}`
-| *Deployment failed* a| `${severity}` `${resource}` `${user}` `${message}`
-| *Deployment success* a|`${severity}` `${resource}` `${user}`
-| *Exceeds event traffic threshold* a| `${severity}` `${resource}`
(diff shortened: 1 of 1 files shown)


ldebello:
I was reading the docs, it is a little confusing at least for me to have a list of "Alert Variables" and later "Variables by Alert" I would have something like

One section explaining the variables availables in all alerts and app alerts, and a list showing only specific variables for alert, for example having "Exceeds event trac threshold" in the list with no specific variables is more confusing because it create a larger list but no add value in have the line, and I would sort the list trying to put together same kind of alerts, for example "CPU usage - CloudHub" is line 1 inside "CloudHub App Alerts" but "Memory usage - CloudHub" in the same category is line 6 and they have the same variables

ldebello (commented):
I have just added a comment in the PR

rodrigo-neira:
Hi @ldebello in our latest ARM sustaining meeting, we discussed that 
_"Number of errors:
The number of errors in one minute reaches the specied limit."_

is not clear. What is considered an **error** here? Can we please have this clarified with the PM and add it to the docs? Let me know if this should be tracked in a separate DOCS/PR. (Check details with Sebastian and Curra) - Thanks!

ldebello:
> Hi @ldebello in our latest ARM sustaining meeting, we discussed that
> _"Number of errors: The number of errors in one minute reaches the specied limit."_
> 
> is not clear. What is considered an **error** here? Can we please have this clarified with the PM and add it to the docs? Let me know if this should be tracked in a separate DOCS/PR. (Check details with Sebastian and Curra) - Thanks!

Hi the number of errors is provided by the runtime, so not sure can be respond that, the mule agent plugin only ask the runtime to get that information and send to the management plane, i am assuming the meaning is the number of flow execution that fails but not sure about that.

ldebello (dismissed):
LGTM

sarathecoubian (approved):
Looks good, but I'm not entirely sure what the bolding/lack of bolding in the modified table is trying to convey.

PR #422, opened 2020-11-18
Added a note about app deployment status

hannanelson:


modules/ROOT/pages/anypoint-platform-cli-commands.adoc
```diff
@@ -1431,6 +1431,9 @@ This command deploys the Mule deployable archive ZIP file that you specify in `<
 You can start typing your application's name and press `tab` for Anypoint Platform CLI to autocomplete it, or you can double tap `tab` for a full list of all the values you can pass. +
 You will have to provide the absolute or relative path to the deployable ZIP file in your local hard drive and the name you give to your application has to be unique.
 
+[NOTE]
+This command's output includes `Status`, which is the application's previous deployment status state.
```
jishun-xue: This command is specifically for new application deployment. So the status that will be returned is **UNDEPLOYED** if the command succeeds which implies the application is uploaded successfully.

Could you update command **modify**'s doc as well about Status in response.

jishun-xue (changes requested):

modules/ROOT/pages/anypoint-platform-cli-commands.adoc
hannanelson: Thanks @jishun-xue - take a look now.

jishun-xue (approved):

jishun-xue (approved):

dukesphere (approved):

PR #437, opened 2021-01-04
Clarified filePath argument

hannanelson:


mcampo (commented):
Users can upload a single-file API (.raml/.yaml/.json) using the CLI, so the conditional is correct.

hannanelson:
CLI bug that prevented uploading other formats has been fixed. Closing.

PR #453, opened 2021-02-01
Update vpn-about.adoc

dtufino:
mentioned PFS requirement as per AWS guidelines


danieldiaz-mule (changes requested):
This is redundant, having a diffie-hellman group in phase 2 implies that PFS is enabled (the group is only required with PFS on).
https://community.cisco.com/t5/vpn/dh-group-in-phase-1-and-phase-2/td-p/2017903

If we want to clarify that PFS is a requirement, we should do it explicitly in the requirements section for each connection type.

hannanelson:
So @danieldiaz-mule, it sounds like this PR is unnecessary and I should close?

danieldiaz-mule:
> So @danieldiaz-mule, it sounds like this PR is unnecessary and I should close?

To me, yes... @dtufino do you have any specific context as to why we need this?

dtufino:
@danieldiaz-mule  For some customers it may not be apparent that PFS is a requirement so we should probably clarify as it is stated in AWS documentation. I agree it should be added in the requirement section instead.

hannanelson:
Closing per Daniel

PR #457, opened 2021-02-09
Update vpn-about.adoc

dtufino:
Reference PFS requirement for IPsec DH groups


danieldiaz-mule (changes requested):
Please change to "Phase 2 Diffie-Hellman groups". Also, please include this requirement for Dynamic connections (as it stands it's only listed for static conns).

danieldiaz-mule (approved):
LGTM

emileemooreVT (approved):

PR #474, opened 2021-03-18
Correction on asset page delete command

constanzaderienzo:
The `asset page delete` command had a comment that was taken from the `asset page download` command which did not apply.


hannanelson (approved):
LGTM!

PR #492, opened 2021-05-06
DOCS-10918 - it's Schedulers on hybrid!

hannanelson:


fpiacentini08 (approved):
LGTM

modules/ROOT/nav.adoc
```diff
@@ -15,7 +15,7 @@
   *** xref:custom-application-alerts.adoc[Custom Application Alerts]
   *** xref:autoscaling-in-cloudhub.adoc[Autoscaling in CloudHub]
  ** xref:managing-applications-on-your-own-servers.adoc[Manage Applications on Your Servers]
-  *** xref:hybrid-schedule-mgmt.adoc[Manage Schedules]
```
fpiacentini08, line 18: Why not renaming the file too?

modules/ROOT/pages/hybrid-schedule-mgmt.adoc
```diff
@@ -32,14 +32,14 @@ Runtime Manager lists the name of each Scheduler component as `polling://{flow_n
 
 Schedule management is useful in the following contexts:
```
fpiacentini08, line 33: Missing "Schedule"

fpiacentini08: No, it's ok. Because it is the schedule of the scheduler. Right?

modules/ROOT/nav.adoc
hannanelson: @fpiacentini08 - We avoid renaming files because there might be links to it. Also, it sounded like the original reporter thought it should be "Schedules" to match CloudHub. If you think it's important, I can make this change.

modules/ROOT/pages/hybrid-schedule-mgmt.adoc
hannanelson: @fpiacentini08 - I'm not sure what you're asking. Should it be Scheduler management?

Mulesoft-Mysti (approved):

modules/ROOT/pages/hybrid-schedule-mgmt.adoc
fpiacentini08: I think it is ok

robertpliu (commented):
Approved.

PR #498, opened 2021-05-24
DOCS-10871 CH VPN manual update

hannanelson:


hannanelson:
Removing - no change to docs needed as this is a temporary change.

PR #503, opened 2021-06-17
DOCS-11182: Added recommendation

hannanelson:


aljrubior (dismissed):
LGTM

modules/ROOT/pages/managing-applications-on-your-own-servers.adoc
```diff
@@ -23,7 +23,8 @@ With applications deployed to your own servers, you can:
 This doc covers deployments on-premises that are deployed both through the https://anypoint.mulesoft.com[cloud based] or the Private Cloud Edition of the Anypoint Platform. Some of the features described in this page are not yet available on the Anypoint Platform Private Cloud Edition. See xref:deployment-strategies.adoc[Deployment Options] for a clear distinction. If you’re interested in requesting the Anypoint Platform Private Cloud Edition, please contact your sales representative.
 
 [NOTE]
-Only the applications deployed through Runtime Manager will be visible and manageable for that tool. If you have applications manually deployed in your on-premise servers, these are not visible in Runtime Manager as the applications are not auto discovered.
+MuleSoft does not support manually deploying apps if you are managing your on-premises servers with Runtime Manager.
+If you manually deploy applications on your on-premise servers, they aren't visible in Runtime Manager as the applications are not auto-discovered.
```
mcurrao-mulesoft: I would add something explicitly stating that they are not only not visible, they are not manageable. Like 

```suggestion
If you manually deploy applications on your on-premise servers, they will not be visible in Runtime Manager (as the applications are not auto-discovered), and you will be unable to manage them through the Anypoint Platform.
```

mcurrao-mulesoft: Also, if names collide for an application manually deployed and an application deployed through the management plane, the manually deployed application *may* be removed and replaced with the new application. But this is not an explicit feature, and we do not provide support for hot-swapping manual deployments for managed deployments AFAIK. So, I am unsure if that should be included. As a warning, perhaps? 

cc: @aljrubior

hannanelson: I think we shouldn't even mention hot-swapping and just leave it as "not supported".

falessio2 (approved):
LGTM

sarathecoubian (approved):

PR #516, opened 2021-07-06
Updated all icon links

Rodgers-Glenn:
Corrected all icon links.

https://www.mulesoft.org/jira/browse/DOCS-11092


dukesphere (approved):
spot checked the links

PR #523, opened 2021-07-13
[ION-9593] Document HTTP Streaming does not work via a S/DLB

dnishant:


modules/ROOT/pages/dedicated-load-balancer-tutorial.adoc
```diff
@@ -43,6 +43,12 @@ Dedicated load balancers are an optional component of Anypoint Platform.
 
 For information about dedicated load balancers, see xref:cloudhub-dedicated-load-balancer.adoc[Dedicated Load Balancers].
 
+---
+
+[IMPORTANT]
+The Loadbalancers do not support HTTP Streaming at this point. When a call is made directly to the worker HTTP streaming behaviour is demonstrated as expected (i.e An error response is received before the full attachment is uploaded)
```
hannanelson:
```suggestion
The DLBs don't support HTTP streaming.
When you make a call directly to the worker, you receive the expected error response before the full attachment is uploaded.
```

modules/ROOT/pages/dedicated-load-balancer-tutorial.adoc
```diff
@@ -43,6 +43,12 @@ Dedicated load balancers are an optional component of Anypoint Platform.
...
+However when a call is made to the Loadbalancer the response is not received until the full upload is complete (this could take many minutes).
```
hannanelson:
```suggestion
However, when you make a call to the DLB, you don't receive the response until the full upload completes, which could take several minutes.
```

nishantkakar:
Should we also add it for SLB here? https://docs.mulesoft.com/runtime-manager/dedicated-load-balancer-tutorial#shared-load-balancers
@hannanelson how do we do this so that it's not duplicating the same infromation?

hannanelson:
@nishantkakar - I think we should put it as a note/important in the load balancer topic: 
https://docs.mulesoft.com/runtime-manager/dedicated-load-balancer-tutorial

> Should we also add it for SLB here? https://docs.mulesoft.com/runtime-manager/dedicated-load-balancer-tutorial#shared-load-balancers
> @hannanelson how do we do this so that it's not duplicating the same infromation?

modules/ROOT/pages/dedicated-load-balancer-tutorial.adoc
```diff
@@ -43,6 +43,11 @@ Dedicated load balancers are an optional component of Anypoint Platform.
 
 For information about dedicated load balancers, see xref:cloudhub-dedicated-load-balancer.adoc[Dedicated Load Balancers].
 
+---
+
+[IMPORTANT]
+Load balancers do not support HTTP Streaming at this point. When a call is made directly to the worker HTTP streaming behaviour is demonstrated as expected (i.e An error response is received before the full attachment is uploaded). However, when you make a call to the load balancer, you don't receive the response until the full upload completes, which could take several minutes.
```
sarathecoubian:
```suggestion
Load balancers do not currently support HTTP streaming. When a call is made directly to the worker, HTTP streaming behavior occurs as expected (that is, an error response is received before the full attachment is uploaded). However, when you make a call to the load balancer, you don't receive the response until the full upload completes, which can take several minutes.
```

sarathecoubian (approved):

PR #542, opened 2021-08-31
DOCS-12057-Fix duplicated `-keypass` and warning about it

mdaloia:
The `-keypass` on 2 commands were repeated generating the following warning:

```
Warning:
The -keypass option is specified multiple times. All except the last one will be ignored.
```

De-duplicated and moved to the end to keep it consistent with the rest of the commands that use:
`-keypass mulesoft -storepass mulesoft`

**Before:**

image

**After:**

image


luanamulesoft:
Created JIRA [DOCS-12057](https://www.mulesoft.org/jira/browse/DOCS-12057) for tracking

luanamulesoft (approved):

PR #586, opened 2022-02-16
W-10587301 ECDSA cipher limitation

luanamulesoft:


shambhuprasad (dismissed):
LGTM

modules/ROOT/pages/lb-cert-validation.adoc
```diff
@@ -306,6 +306,30 @@ Run this command against the DLB SSL endpoint:
 * If the command returns `curl: (35) Server aborted the SSL handshake`, TLS 1.0 is not supported.
 --
 
+== Limitation
+
+To use ECDSA based authentication cipher, you must generate the certificate and key pair using ECDSA based algorithm and apply it to the SSL certificate and key configuration on the DLB.
```
kbostrom: GLOBALLY include a hyphen before "based" in such compound constructions. 
Eliminate modal verbs like "must" and "can" when it does not impair the intended meaning to do so.
Reread sentences that include "it" to ensure that it is very clear to what "it" refers, or reword just to make sure; I was not sure if "it" in the following sentence refers to "cipher," "certificate and pair," or "algorithm" and so I took a guess (please verify).
```suggestion
To use an ECDSA-based authentication cipher, generate the certificate and key pair using an ECDSA-based algorithm and apply the same algorithm to the SSL certificate and key configuration on the DLB.
```

modules/ROOT/pages/lb-cert-validation.adoc
```diff
@@ -306,6 +306,30 @@ Run this command against the DLB SSL endpoint:
...
+
+If you try to use ECDSA based authentication cipher with RSA based certificate and key pair, the SSL handshake will fail and you will receive a either of the following error messages:
```
kbostrom: Ditto prev. comment about hyphenating "based" compounds.

Also, we use an article before any proper noun (ECDSA, RSA) when it is used as an adjective (ECDSA-based, RSA-based) followed by a common noun (cipher, pair): a cipher, a pair. Let me know if you would like me to explain this further.

Finally, GLOBALLY search for future-tense words like "will" and reword to present tense unless it is vital to focus on a future functionality rather than an ongoing (present) functionality. It's best to imply that something always works one way rather than to imply that this is only in the future.
```suggestion
If you try to use an ECDSA-based authentication cipher with an RSA-based certificate and key pair, the SSL handshake fails and you receive either of the following error messages:
```

modules/ROOT/pages/lb-cert-validation.adoc
```diff
@@ -306,6 +306,30 @@ Run this command against the DLB SSL endpoint:
...
+----
+
+or
+
+[source,linenums]
+----
+Error: write EPROTO 140434219616616:error:10000410:SSL routines:OPENSSL_internal:SSLV3_ALERT_HANDSHAKE_FAILURE:../../third_party/boringssl/src/ssl/tls_record.cc:594:SSL alert number 40 140434219616616:error:1000009a:SSL routines:OPENSSL_internal:HANDSHAKE_FAILURE_ON_CLIENT_HELLO:../../third_party/boringssl/src/ssl/handshake.cc:603:
+----
+
+Current DLB configuration only allows for one type of key pair for an SSL endpoint. You can configure the DLB with SSL certificate and key that are generated either with RSA or ECDSA algorithm but not both.
```
kbostrom:
```suggestion
Current DLB configuration permits only one type of key pair for an SSL endpoint. You can configure the DLB to use an SSL certificate and a key that are generated using either an RSA or ECDSA algorithm, but not both.
```
See also alternative wording mentioned below.

modules/ROOT/pages/lb-cert-validation.adoc
```diff
@@ -306,6 +306,30 @@ Run this command against the DLB SSL endpoint:
...
+
+[NOTE]
+If you configure a DLB with an ECDSA based key pair, clients using RSA based ciphers won't be able to connect to that DLB.
```
kbostrom: Is this already implied by the previous text: do we have to clarify in this way?
Also, does this work in reverse, in which ECDSA clients can't connect to RSA pairs?

Perhaps Line 327 should be reworded to say...: 

Because current DLB configuration permits only one type of SSL certificate and key pair per SSL endpoint, a client using an RSA-based cipher can not connect to an ECDSA-based configuration, and vice versa.

kbostrom (dismissed):
CE/CU

kbostrom (approved):
proof

